PR #4, opened 2023-03-11
Create Avatar component

gsouquet:
Screenshot 2023-03-11 at 11 19 06

@janogarcia would you be able to provide me with 8 color codes for the username mapping?

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -1,0 +1,67 @@
+/*
+Copyright 2023  New Vector Ltd
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React, { Suspense } from "react";
+import { getInitialLetter } from "../../utils/string";
+import { SuspenseImg } from "../../utils/SuspenseImg";
+import styles from "./Avatar.module.css";
+import { useUsernameColor } from "./useUsernameColor";
+
+type AvatarProps = {
+  src?: string;
+  id: string;
+  name: string;
+  type?: "square" | "round";
+  size?: CSSStyleDeclaration["height"];
+};
+
+export const Avatar = ({
+  src,
+  id,
+  name = "",
+  type = "round",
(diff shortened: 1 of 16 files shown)


src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,74 @@
...
+        src={src}
+        className={classnames(styles.avatar, className)}
+        data-type={type}
+        style={style}
+        width={size}
+        height={size}
+        title={id}
+      />
+    </Suspense>
+  );
```
justjanne, line 73: This is good, but I feel like an early-return style would flow better here, especially if we get more failure conditions in the future. That said, this is just personal preference, not an actual request for you to change this.

```suggestion
  if (!src) {
    return imagelessAvatar;
  }

  return (
    <Suspense fallback={imagelessAvatar}>
      <SuspenseImg
        src={src}
        className={classnames(styles.avatar, className)}
        data-type={type}
        style={style}
        width={size}
        height={size}
        title={id}
      />
    </Suspense>
  );
```

src/utils/SuspenseImg.tsx
```diff
@@ -0,0 +1,66 @@
...
+          this.__cache.set(src, true);
+        })
+      );
+    }
+    if (this.__cache.get(src) instanceof Promise) {
+      throw this.__cache.get(src);
+    }
+    return this.__cache.get(src) as boolean;
+  },
+};
```
justjanne, line 43: As this considers an image, once it has loaded, cached forever, it might give false positives for images that have been evicted by the browser. On the other hand it might lead to a flash of the fallback image if the browser still has an image cached from a previous session.

Fine for now, but in the future, it might be ideal if we implement this in a better way.

src/components/Avatar/useUsernameColor.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Determines a color to assign to a given username
+ * @param id - a Matrix username ID
+ * @returns a hexadecimal color
+ */
+export function useUsernameColor(id: string): number {
+  const MIN = 1;
+  const MAX = 8;
+  // Sum up the values of all the char codes in the string
+  const charCodeSum = id.split("").reduce((sum, char) => {
+    return sum + char.charCodeAt(0);
```
justjanne, line 27: Note that for names containing unicode characters outside the basic multilingual plane this will use the surrogate pairs. It relies on the specific UCS-2 16-bit encoding common to JS/Java/Windows/Qt

justjanne (approved):

src/components/Avatar/useUsernameColor.ts
gsouquet: As discussed with Janne, most of our codebase will use `utf-8` and that shouldn't present a great threat.
There's also a chance that those colors would go away altogether. 

We're punting this decision to a later stage

PR #5, opened 2023-03-15
Create basic button component

gsouquet:


janogarcia (commented):
I've made some notes on the tokens that would address those gaps. Actually some of those are in the works. For example, for `.button[data-kind="primary"] { color: #fff; }` most probably we would use something like `color.text.on-primary`.

Is there anything else I should look into in this PR?

Thanks!

janogarcia:
Let me know if I should self-request a new review and approve it, or just go ahead and approve it if you don't need me to review anything else. 👍

nadonomy (approved):
LGTM to merge and iterate later on any values we want to tweak.

PR #6, opened 2023-03-20
Create Alert component

gsouquet:
Following https://www.figma.com/file/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?node-id=101%3A16968&t=ITmcXhLdT4wQRu3j-1

There are bunch of semantic tokens that need to be defined (FAO @janogarcia )
I am also using the thread icon as a placeholder. This needs to be changed before merging this


src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,52 @@
...
+  onClose,
+  ...props
+}: PropsWithChildren<AlertProps<C>>): JSX.Element => {
+  const Component = as || "div";
+  const classes = classNames(styles.alert, className);
+
+  return (
+    <Component {...props} className={classes} data-type={type} tabIndex={0}>
+      {/* TODO: Replace icon thread with the close icon */}
+      <Icon icon="thread" size={40} />
```
justjanne, line 43: This should be an info/warn/danger icon, right? Or do we really want two close icons on the same alert?

justjanne (approved):

src/components/Alert/Alert.tsx
gsouquet: The comment is incorrect, you're right!

kerryarchibald, line 43: This seems too big compared to the design

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export const Alert = <C extends React.ElementType = "div">({
+  as,
+  type,
+  children,
+  className,
+  onClose,
+  ...props
+}: PropsWithChildren<AlertProps<C>>): JSX.Element => {
+  const Component = as || "div";
```
kerryarchibald, line 37: What do you see as the use case for this? Would it be good to lock it down to a subset of elements that make sense?

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <Component {...props} className={classes} data-type={type} tabIndex={0}>
+      {/* TODO: Replace icon thread with the close icon */}
+      <Icon icon="thread" size={40} />
+      <div className={styles.content}>{children}</div>
+      {/* TODO: Setup an i18n function for the aria label below */}
+      {/* TODO: Replace icon thread with the close icon */}
+      {onClose && (
+        <Icon icon="thread" size={16} onClick={onClose} aria-label="Close" />
```
kerryarchibald, line 48: I think this should be wrapped in a button

src/components/Alert/Alert.module.css
```diff
@@ -0,0 +1,105 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/* TODO: Review entire file for semantic token definiton */
+
+.alert {
+  display: flex;
+  align-items: center;
```
kerryarchibald, line 21: (More of a design comment, but...) Vertically centering the close button means it won't be in a consistent position (depending on longer/shorter content). This would be super annoying if these are used as dialogs

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,52 @@
...
+import styles from "./Alert.module.css";
+
+type AlertProps<C extends React.ElementType> = {
+  as?: C;
+  type: "success" | "critical" | "info";
+  className?: string;
+  onClose?: (e: React.MouseEvent) => void;
+} & React.ComponentPropsWithoutRef<C>;
+
+export const Alert = <C extends React.ElementType = "div">({
```
kerryarchibald, line 29: What do you think about allowing HTML attributes in the props? I'm specifically thinking about testIds

src/components/Alert/Alert.module.css
```diff
@@ -0,0 +1,105 @@
...
+  gap: var(--cpd-space-3x);
+  padding: var(--cpd-space-4x);
+  border-radius: 12px;
+  border: 1px solid;
+}
+
+.alert[data-type="success"] {
+  color: var(--cpd-color-green-600);
+  background-color: var(--cpd-color-green-200);
```
kerryarchibald, line 30: Are these colours back to front? light text on dark instead of dark text on light bg?

gsouquet: They are back to front yes, but we should never use these colour scales directly.
Jano is defining semantic tokens, and we should consume those... but they were not defined at the time of creating this PR.

src/components/Alert/Alert.tsx
gsouquet: It does accept HTML attributes. In the props definition the `React.ComponentPropsWithoutRef<C>` part does just that

src/components/Alert/Alert.module.css
gsouquet: Makes a lot of sense! Will update that

src/components/Alert/Alert.tsx
gsouquet: You're probably right, reading through https://developer.mozilla.org/en-US/docs/Web/HTML/Element#text_content the two elements that make proper sense for this would be a `div` and ` p`.

But if the choices are that limited, wouldn't we be better off by only allowing paragraphs?

kerryarchibald: I personally lean more towards locking things down, and only adding customisation like this when a real use-case presents itself.
There's bigger question of where we want this library to be - fully prescriptive <--> fully customisable. Not sure if that's been documented/discussed anywhere, also maybe a problem for future us? (Not trying to derail this PR!)

PR #7, opened 2023-03-21
Create compound forms structure

gsouquet:
This pull request lays the foundation for forms in Compound.

It creates a thin wrapper around [Radix UI Forms](https://www.radix-ui.com/docs/primitives/components/form) and aims at re-using the primitives it exposes but also set the UI we want as well as the a11y attributes we're interested in

Figma spec: https://www.figma.com/file/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?node-id=101-16968&t=ld4mSXISz6Jt6t6f-0

This is a work in progress, here's a non comprehensive list of things that are not yet tackled

- The message components awaiting implementation of https://github.com/vector-im/compound-web/pull/6
- Only input controls are supported
- Missing the whole spacing and layout of components when they're living next to each other

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@vector-im/compound-design-tokens",
+  "name": "@vector-im/compound-web",
   "version": "0.0.1",
   "description": "Compound components for the Web",
   "type": "module",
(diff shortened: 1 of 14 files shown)


alunturner (approved):
Looks good to me, appreciate this is a work in progress so the comment doesn't need any actions

PR #10, opened 2023-03-22
Create ActionControl component

gsouquet:
Depends on and includes the diff of https://github.com/vector-im/compound-web/pull/11
Also fixes https://github.com/vector-im/compound/issues/76

- Create ActionControl component
- Create password reveal component

https://user-images.githubusercontent.com/769871/227013455-d2eec7e4-d0e5-463d-b4e7-b93e9a43d46b.mov


src/components/Form/Controls/Password.tsx
```diff
@@ -0,0 +1,47 @@
...
+ */
+export function PasswordControl(
+  props: PropsWithChildren<React.ComponentProps<typeof Control>>
+): JSX.Element {
+  const [label, setLabel] = useState<"show" | "hide">("show");
+  const isHidden = label === "show";
+
+  const onActionClick = useCallback(() => {
+    setLabel(isHidden ? "hide" : "show");
+  }, [label]);
```
alunturner, line 35: Is the expectation that this will grow to have more than two possible labels? 

If not, then I quite like using useReducer to make a toggle something like:
`const [isHidden, toggleIsHidden] = useReducer(state => !state, true)`.

Not really useful if this will be extended past two possible labels of course.

gsouquet: Really good shout, that simplified the approach quite a bit!

src/components/ActionControl/ActionControl.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <ActionControl
+        Icon={ThreadIcon}
+        actionLabel="Click me!"
+        onActionClick={spy}
+      />
+    );
+
+    const actionBtn = getByLabelText(container, "Click me!");
+
+    fireEvent.click(actionBtn);
```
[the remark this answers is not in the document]
gsouquet: Good shout 👍 I've tracked that as part of https://github.com/vector-im/compound/issues/78

I'm not sure I see the value the difference between `screen` and `container` will bring here?
Regarding setting up guidelines, I would prefer to capitalise on the ones set as part of Element Web as much as possible to have consistent guidelines throughout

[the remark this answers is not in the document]
alunturner: I think the main benefit is not ever having to change imports/what you've destructured from the call to `render`. Consistency is probably more important though, so lining up with Element Web makes sense.

alunturner (approved):

janogarcia:
@gsouquet The live demo in the test deployment doesn't seem to work for me. Clicking on the icon won't switch between masked and unmasked.

src/components/ActionControl/ActionControl.module.css
```diff
@@ -0,0 +1,37 @@
...
+limitations under the License.
+*/
+
+.actioncontrol {
+  max-width: fit-content;
+  position: relative;
+}
+
+input.input {
+  padding-inline-end: (--cpd-space-12x);
```
kerryarchibald, line 23: missing `var()`?

src/components/ActionControl/ActionControl.tsx
```diff
@@ -0,0 +1,67 @@
...
+  Icon,
+  className,
+  actionLabel,
+  onActionClick,
+  ...props
+}: PropsWithChildren<ActionControlProps>) => {
+  // TODO: Update the class name to something related to the component name
+  const classes = classnames(styles.actioncontrol, className);
+  // TODO: I don't know how to type this properly
+  const IconComp = Icon as typeof React.Component;
```
kerryarchibald, line 42: Can you use the type from the '*.svg' module? `const ReactComponent: React.FunctionComponent<
    React.SVGAttributes<SVGElement>
  >;`

gsouquet: Yup, that was the old way I had created icons.
Your suggestion is defo the way to go!

PR #13, opened 2023-03-24
Upload coverage artifacts and configure sonarcloud

gsouquet:
Fixes https://github.com/vector-im/compound/issues/79


sonar-project.properties
```diff
@@ -0,0 +1,10 @@
+sonar.projectKey=compound-web
```
t3chguy, line 1:
```suggestion
sonar.projectKey=vector-im_compound-web
```

t3chguy: image

sonar-project.properties
```diff
@@ -0,0 +1,10 @@
...
+sonar.organization=vector-im
+
+
+# sonar.sources=src,res
+# sonar.tests=test,cypress
+sonar.exclusions=__test__,docs
+
+sonar.typescript.tsconfigPath=./tsconfig.json
+sonar.javascript.lcov.reportPaths=coverage/lcov.info
```
t3chguy, line 10: this seems unused given the above coverage artifact is uploaded but never downloaded

.github/workflows/pull_request.yml
```diff
@@ -26,5 +26,17 @@ jobs:
...
+              with:
+                  name: coverage
+                  path: |
+                      coverage
+                      !coverage/lcov-report
+            - name: SonarCloud Scan
+              uses: SonarSource/sonarcloud-github-action@master
+              env:
+                GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                SONAR_TOKEN: ${{ secrets.SONARCLOUD_TOKEN }}
```
t3chguy, line 42: This won't work for forked PRs, thus will mean you can't use coverage as a branch protection

t3chguy: https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

gsouquet: Makes sense! I'm happy to punt that for now as we don't expect contributions for now.
But I have tracked it as part of https://github.com/vector-im/compound/issues/80

t3chguy (approved):

PR #20, opened 2023-04-26
Create Checkbox component

gsouquet:

PR #21, opened 2023-04-26
Set correct storybook canvas bg color

gsouquet:
Requires https://github.com/vector-im/compound-design-tokens/pull/45

Fixes the issue raised by @t3chguy in https://github.com/vector-im/compound-web/pull/20

- Set correct storybook canvas bg color
- Add theme selector dropdown

Screenshot 2023-04-26 at 14 18 56

### DIFF
--- a/.storybook/preview.ts
+++ b/.storybook/preview.ts
@@ -9,4 +9,5 @@ export const parameters = {
       date: /Date$/,
     },
   },
+  backgrounds: { disable: true },
 };
--- a/.storybook/preview.ts
+++ b/.storybook/preview.ts
@@ -10,4 +10,21 @@ export const parameters = {
     },
   },
   backgrounds: { disable: true },
+  themes: {
+    default: "light",
+    list: [
+      { name: "light", class: "cpd-theme-light", color: "#fff" },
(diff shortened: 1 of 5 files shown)


t3chguy:
The bg isn't changing for me

image

gsouquet:
Yes, it requires vector-im/compound-design-tokens#45 to land first. Got tricked by CSS specificity unfortunately!

.storybook/preview.ts
```diff
@@ -9,4 +9,22 @@ export const parameters = {
...
+      { name: "dark", class: "cpd-theme-dark", color: "#101317" },
+      {
+        name: "light-high-contrast",
+        class: "cpd-theme-light-hc",
+        color: "#fff",
+      },
+      {
+        name: "dark-high-contrast",
+        class: "cpd-theme-dark-hc",
+        color: "#101317",
```
alunturner, line 26: Probably a silly question - how come we have to hardcode these hex values?

alunturner (approved):

.storybook/preview.ts
gsouquet: We don't have access to the compound design tokens there. Which is why we had to hardcode them

PR #24, opened 2023-05-03
Integrate with OIDC codebase

sandhose:
This pull request does a couple of things in order to improve the integration with the OIDC codebase

- Exposes TypeScript types
- Lets the package be imported via a GitHub URL rather than an npm install
- Lets you import the CSS file in other projects


gsouquet (approved):
That looks great to me, thank you for looking into this 🚀

PR #25, opened 2023-05-04
Create Toggle component

gsouquet:
[Figma reference](https://www.figma.com/file/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?type=design&node-id=129%3A4463&t=Jfso4ZUsWY6keIdq-1)


src/components/Toggle/Toggle.module.css
```diff
@@ -0,0 +1,91 @@
...
+  outline: 2px solid var(--cpd-color-border-focused);
+  outline-offset: 1px;
+}
+
+:checked + .toggle-ui::after {
+  transform: translateX(100%);
+  background: var(--cpd-color-icon-on-solid-primary);
+}
+
+[type="checkbox"][disabled] .toggle-ui {
```
kerryarchibald, line 80: I think these rules need a `+`

src/components/Toggle/Toggle.module.css
```diff
@@ -0,0 +1,91 @@
...
+  grid-area: 1/1;
+  width: 44px;
+  height: 24px;
+  box-sizing: border-box;
+}
+
+.toggle > input[type="checkbox"] {
+  opacity: 0;
+  margin: 0;
+  cursor: pointer;
```
kerryarchibald, line 33: The disabled toggle still uses a pointer cursor

kerryarchibald (commented):
Looks good!
Couple of small nitpicks about disabled state

nadonomy:
Will defer component review to @janogarcia - but @gsouquet is there a strong rationale for calling this a toggle instead of a 'checkbox' per the specs?

Colloquially, we often use toggles to describe [toggle switches](https://i.pinimg.com/originals/3d/16/f5/3d16f5908ad93653b0292dbdcd039067.png) on mobile, so I think checkbox is more descriptive.

janogarcia:
@gsouquet As @nadonomy pointed put the Figma link in the description is pointing to the Checkbox spec, while the implementation seems to be that of the Toggle component.

kerryarchibald (approved):

janogarcia (commented):
**Storybook**

- The embedded Figma frame is pointing to the Checkbox spec instead of the Toggle component.
- We are missing the disabled state, so I had to revert to manually adding the `disabled ` attribute via devtools.
- Could we add the labelled components as well (label with start/end position options, optional help text), or is that out of the scope for this milestone?

**UI specs**

- When in the `checked` state, the border color should be `transparent` so that we have a continuous solid color for the background. Same for the `checked` + `hovered ` state.
- The `checked` + `disabled ` state is not rendered correctly, it results in a plain gray surface. The token mapping for `[type="checkbox"][disabled] + .toggle-ui::after` is wrong. Also, the border should be transparent, as noted above.
- Default to `ease` instead of `ease-in-out` for snappier interactions. Note: We haven't formalized the motion guidelines or any related tokens, but we could take this as an opportunity to formalize the first one.

**Accessibility: Focus state**

We don't need to show the focus state on checkbox/radio/toggle component when the input doesn't come from keyboard navigation. 

Can we reliably display it only if the input event comes from a keyboard but not for example when clicking with a pointer (mouse/trackpad)?

**Accessibility: Scalable dimensions**

I'd suggest that sizing/dimensions are defined as `rem` instead of `px`, that way the components will adapt gracefully to the user's system font size, in a similar way to changing the font size at the system level in iOS/Android.

A few exceptions may apply, like fixed width borders (e.g., we want to keep a border at`1px` no matter the system font size)

janogarcia (commented):
The PR comment above got double posted. I've deleted the contents for this one.

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,39 @@
...
+import React, { PropsWithChildren } from "react";
+import styles from "./Toggle.module.css";
+
+type ToggleProps = {
+  className?: string;
+} & React.ComponentPropsWithoutRef<"input">;
+
+/**
+ * Toggle form control
+ */
```
richvdh, line 27: I might be misunderstanding something here but personally I'd expect to see more information about what exactly this is ("React component which renders a checkbox control for a form")  and how it can be used by an application:

What are the supported properties? Examples of common ones (`disabled`, `checked`) would be helpful. How does the application read the state?

gsouquet: I agree, the description of the component could be much better here! I will update that. 

I am not quite sure we should write down the supported properties here, they're defined by the TypeScript type just above, and would be concerned these two things go out of sync.

The discovery of the component variants will be achieved through Storybook stories in the long term, https://compound.element.io/?path=/story/compound-web_alert--success (see the right panel as an example)
But we are yet to formalise an approach here

src/components/Toggle/Toggle.module.css
```diff
@@ -0,0 +1,95 @@
...
+
+.toggle-ui {
+  pointer-events: none;
+  width: 44px;
+  aspect-ratio: 1 / 1;
+  border-radius: 9999px;
+  border: 1px solid var(--cpd-color-border-interactive-primary);
+  position: relative;
+  padding: 1px;
+  transition: all 0.2s ease-in-out;
```
gsouquet, line 48: From @janogarcia 

> Default to ease instead of ease-in-out for snappier interactions. Note: We haven't formalized the motion guidelines or any related tokens, but we could take this as an opportunity to formalize the first one.

```suggestion
  transition: all 0.2s ease;
```

src/components/Toggle/Toggle.module.css
```diff
@@ -0,0 +1,95 @@
...
+  background: var(--cpd-color-bg-canvas-disabled);
+}
+
+[type="checkbox"][disabled] + .toggle-ui::after {
+  background: var(--cpd-color-bg-action-primary-disabled);
+}
+
+[type="checkbox"][disabled]:checked + .toggle-ui {
+  background: var(--cpd-color-bg-action-primary-disabled);
+}
```
gsouquet, line 95: From @janogarcia 

> The checked + disabled state is not rendered correctly, it results in a plain gray surface. The token mapping for [type="checkbox"][disabled] + .toggle-ui::after is wrong. Also, the border should be transparent, as noted above.

src/components/Toggle/Toggle.module.css
```diff
@@ -0,0 +1,95 @@
...
+  cursor: pointer;
+}
+
+.toggle > input[type="checkbox"][disabled] {
+  cursor: not-allowed;
+}
+
+.toggle-ui {
+  pointer-events: none;
+  width: 44px;
```
gsouquet, line 42: From @janogarcia 

> I'd suggest that sizing/dimensions are defined as rem instead of px, that way the components will adapt gracefully to the user's system font size, in a similar way to changing the font size at the system level in iOS/Android.
A few exceptions may apply, like fixed width borders (e.g., we want to keep a border at1px no matter the system font size)

src/components/Toggle/Toggle.module.css
```diff
@@ -0,0 +1,95 @@
...
+  border-radius: 50%;
+  background: var(--cpd-color-bg-action-primary-rest);
+  transform: translateX(0);
+  transition: all 0.2s ease-in-out;
+}
+
+[type="checkbox"]:focus + .toggle-ui {
+  outline: 2px solid var(--cpd-color-border-focused);
+  outline-offset: 1px;
+}
```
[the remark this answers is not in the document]
gsouquet: Yes, i've added a `onMouseDown` handler to deal with that scenario.

src/components/Toggle/Toggle.module.css
gsouquet: Done in this context, there are still a few places in the codebase that would benefit from this change.

src/components/Toggle/Toggle.module.css
gsouquet: The background should not be transparent, according to the design it is the same color as the background. I've made the adjustment! 

Thank you for the heads up on the visual glitch!

PR #37, opened 2023-06-19
Do not inline the assets in the built package

sandhose:
When I started using compound, I figured that the output CSS was over 1Mb. This was due to the assets being inlined.

This PR opts out of Vite's library mode. The downside is that we're not building a CommonJS-compatible bundle, and that we have to keep track of all external dependencies, but the upside is that the bundlers we use in the final apps should be a lot more intelligent about assets.

I also noticed that in the output, it was using a mix of `React.createElement` and the new `_jsx` runtime. This was due to svgr handling the JSX transformation itself, so I made sure it was opting in the new JSX runtime.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -6,38 +6,78 @@ import dts from "vite-plugin-dts";
 import svgr from "vite-plugin-svgr";
 
 export default defineConfig({
-  publicDir: "res",
   build: {
     target: browserslistToEsbuild(),
     assetsInlineLimit: 0, // To disable assets inlining as base64 in the bundle
-    lib: {
-      entry: resolve(__dirname, "src/index.ts"),
-      name: "@vector-im/compound-web",
-      fileName: "compound-web",
-    },
+
+    // We expect to be used as a dependency with a bundler, so we don't need to
+    // minify and let the bundler deal with the sourcemaps
+    sourcemap: true,
+    minify: false,
+    cssMinify: false,
+
     rollupOptions: {
-      external: ["react", "react-dom"],
+      // We are *not* running in library mode because inlines all the assets.
+      // Instead, we're running in normal mode & specifying the entrypoint here.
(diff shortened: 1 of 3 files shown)


vite.config.ts
```diff
@@ -6,38 +6,77 @@ import dts from "vite-plugin-dts";
...
+    // We expect to be used as a dependency with a bundler, so we don't need to
+    // minify and let the bundler deal with the sourcemaps
+    sourcemap: true,
+    minify: false,
+    cssMinify: false,
+
     rollupOptions: {
-      external: ["react", "react-dom"],
+      // We are *not* running in library mode because inlines all the assets.
+      // Instead, we're running in normal mode & specifying the entrypoint here.
```
gsouquet, line 21: Let's add a link to the vite issue that tracks this, https://github.com/vitejs/vite/issues/3295

vite.config.ts
```diff
@@ -6,38 +6,77 @@ import dts from "vite-plugin-dts";
...
+      // Must be in sync with the list of dependencies in the package.json
+      // It includes all dependencies except @vector-im/compound-design-tokens
+      external: [
+        "react",
+        "react-dom",
+        "react/jsx-runtime",
+        "lodash",
+        "classnames",
+        "@radix-ui/react-form",
+      ],
```
gsouquet, line 33: Could we build that list dynamically by reading the `package.json` and omitting `@vector-im/compound-design-tokens` from that list.

I'm pretty certain we will fail to keep those two things in sync otherwise.

gsouquet (changes requested):
That looks pretty great! I had this on the backburner, very much appreciate you taking the time to look into it.

vite.config.ts
sandhose: Done

gsouquet (approved):

PR #38, opened 2023-06-21
Add Cloudflare Pages headers file

laurencegill:
Disable the default behaviour of adding access-control-allow-origin: * to pages sites


germain-gg (approved):
Looks good to me!

PR #39, opened 2023-06-21
Create StackedAvatars component

gsouquet:
Screenshot 2023-06-21 at 10 16 25


src/components/Avatar/StackedAvatars.tsx
```diff
@@ -0,0 +1,56 @@
...
+  return (
+    <div className={classnames(styles["stacked-avatars"], className)}>
+      {avatars.map((avatar) => (
+        <Avatar key={avatar.id} {...avatar} size={size} type="round" />
+      ))}
+      {/* Adds the SVG to the document if it's not been added already
+          It is then used via `clip-path` in CSS 
+      */}
+      {!document.getElementById(AVATAR_MASK_ID) &&
+        createPortal(<AvatarClipPath id={AVATAR_MASK_ID} />, document.body)}
```
florianduros, line 53: I think it'll be great to have some explanations:
- why we need to have a portal here
- what is the purpose of `AvatarClipPath` here

florianduros (changes requested):

src/components/Avatar/StackedAvatars.tsx
gsouquet: Does the comment above not address those comments? I can expand on that, but the sole purpose is so that we can use it with `clip-path` 🤔

t3chguy (dismissed):
The size prop seems to not work. If you specify anything other than 64px it fails. 64 and 65 seem to work, but 200 and 66 do not

image
image

gsouquet:
@t3chguy i don't know how you've been testing nor which browsers. On Firefox + Chrome I get the proper rendering working for any value from 16px and above

Could you elaborate and give reproduction steps?

Screenshot 2023-06-21 at 10 48 23
Screenshot 2023-06-21 at 10 48 30

t3chguy:
image

Stock Firefox 114.0.1 (64-bit) on macOS

florianduros (approved):
LGTM

PR #41, opened 2023-06-28
Refactor useUsernameColor to useIdColorHash

germain-gg:
The hook naming and documentation drifted away from its actual implementation.
This brings it closer to what it actually does!

### DIFF
--- a/src/components/Avatar/useIdColorHash.ts
+++ b/src/components/Avatar/useIdColorHash.ts
@@ -15,11 +15,12 @@ limitations under the License.
 */
 
 /**
- * Determines a color to assign to a given username
- * @param id - a Matrix username ID
- * @returns a hexadecimal color
+ * Determines a number for a given Matrix ID, helps disambiguating users
+ * who are trying to impersonate someone else.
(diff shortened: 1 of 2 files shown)


src/components/Avatar/useIdColorHash.ts
```diff
@@ -15,11 +15,12 @@ limitations under the License.
...
- * Determines a color to assign to a given username
- * @param id - a Matrix username ID
- * @returns a hexadecimal color
+ * Determines a number for a given Matrix ID, helps disambiguating users
+ * who are trying to impersonate someone else.
+ * @param id - a Matrix ID
+ * @returns a hash of the Matrix ID
  */
-export function useUsernameColor(id: string): number {
+export function useIdColorHash(id: string): number {
```
weeman1337, line 23: Does it only apply to MXIDs or can it also be used for other Ids, such as room Ids? Just asking because I wonder if this and Avatar should be a bite more generic to cover rooms as well…

weeman1337 (approved):
Thanks for updating the comment :+1:

PR #46, opened 2023-08-01
Fix React 17/Webpack 4 compatibility

sandhose:
- Move react* to peer dependencies, making it compatible with React 17
- Stop using React.JSX.Element and explicitly type all components
- Add react and react-dom to the dev dependencies
- Upgrade typescript and vite-plugin-dts to fix the output files


src/components/ActionControl/__snapshots__/ActionControl.test.tsx.snap
```diff
@@ -1,32 +1,46 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
-exports[`ActionControl renders 1`] = `
+exports[`ActionControl > renders 1`] = `
 <DocumentFragment>
   <form
-    class="root"
+    class="_root_dc87d2"
```
germain-gg, line 6: Do you know if the suffix is deterministic or will it change or every build?
We need to find a solution if it will change for every build

germain-gg (commented):
Overall looks good to me! As soon as we can get a confirmation with the snapshot test id be happy to go with this!

src/components/ActionControl/__snapshots__/ActionControl.test.tsx.snap
sandhose: it is generated from the filename so it should be stable: https://vitest.dev/config/#css-modules-classnamestrategy

germain-gg: Seems scoped on the filename, which is a stable identifier, https://vitest.dev/config/#css-modules-classnamestrategy

germain-gg (approved):

PR #47, opened 2023-08-01
Forward refs to most components

sandhose:
This makes it so that most components properly forward their ref to the underlying DOM node.
I needed this on the Form.Root element mostly, to be able to clear it, but I extended it to most interactive components.

One notable exception is the `<Button>` component (and therefore the `<Form.Submit>` component), because it is a mess to type correctly because of the generic type on it


germain-gg (approved):
Looks great! Thank you

PR #48, opened 2023-08-01
Stop embedding the Inter font and publish CommonJS packages again

sandhose:
This stops the embedding of the inter font and instead rely on the @fontsource/inter package.

This means that consumers need to install the package and import the font themselves.
It is set as peer dependency, so it will warn if the consumer does not have it installed.

It also means we can switch back to library mode, and emit commonjs builds alongside esm builds.

For consumers, the change is instead of importing just `@vector-im/compound-web/dist/index.css`, they need to:

```css
@import url("@fontsource/inter/400.css");
@import url("@fontsource/inter/500.css");
@import url("@fontsource/inter/600.css");
@import url("@fontsource/inter/700.css");
@import url("@vector-im/compound-design-tokens/assets/web/css/compound-design-tokens.css");
@import url("@vector-im/compound-web/dist/style.css");
```


src/styles/global.css
```diff
@@ -15,51 +15,10 @@ See the License for the specific language governing permissions and
...
 
-/* Inter unexpectedly contains various codepoints which collide with emoji, even
-   when variation-16 is applied to request the emoji variant.  From eyeballing
-   the emoji picker, these are: 20e3, 23cf, 24c2, 25a0-25c1, 2665, 2764, 2b06, 2b1c.
-   Therefore we define a unicode-range to load which excludes the glyphs
-   (to avoid having to maintain a fork of Inter). */
 :root {
-  /* This is a false positive */
-  /* stylelint-disable unit-no-unknown */
-  --cpd-inter-unicode-range: U+0000-20e2, U+20e4-23ce, U+23d0-24c1, U+24c3-259f,
```
sandhose, line 26: I just don't know about that. I know the @font-face in @fontsource/inter has some unicode ranges, but I don't know if it properly excludes the emojis or not. Do we have a way to test that?

germain-gg: So i've actually removed `global.css` from the exported CSS. 
It's still used in Storybook, but right now it is not a great idea to have all of that.

This bit is particularly critical for Element Web, but a bit less for other projects. We do have visual regression tests for it on Percy yes. But only for EW.

package.json
```diff
@@ -98,13 +102,14 @@
...
     "@radix-ui/react-form": "^0.0.3",
-    "@vector-im/compound-design-tokens": "0.0.3",
     "classnames": "^2.3.2",
     "lodash": "^4.17.21"
   },
   "peerDependencies": {
+    "@fontsource/inter": "^5",
     "@types/react": "*",
     "@types/react-dom": "*",
+    "@vector-im/compound-design-tokens": "0.0.3",
```
germain-gg, line 112:
```suggestion
    "@vector-im/compound-design-tokens": "^0.0.3",
```

germain-gg (approved):
Makes sense, looks good to me!

PR #49, opened 2023-08-02
Clean up the build

sandhose:
- in case we're building with the jsx-dev-runtime, it doesn't get included anymore
 - react-dom/server doesn't get included anymore
 - ensure we're not relying on babel
 - make sure we're not using the React default export in the resulting bundle

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -28,7 +28,9 @@ export default defineConfig({
       external: [
         "react",
         "react-dom",
+        "react-dom/server",
         "react/jsx-runtime",
+        "react/jsx-dev-runtime",
         "lodash",
         "classnames",
         "@radix-ui/react-form",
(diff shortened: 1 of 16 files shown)


robintown (approved):

vite.config.ts
```diff
@@ -28,7 +28,9 @@ export default defineConfig({
       external: [
         "react",
         "react-dom",
+        "react-dom/server",
         "react/jsx-runtime",
+        "react/jsx-dev-runtime",
```
germain-gg, line 33: Why? I understand react dom. But not sure I have the rationale behind this one? Is that something that usually gets injected by webpack dev server or similar?

germain-gg (approved):
Looks good to me.
In the short period of time that this package lived we've had a couple of issues related to bundle size. I'll see if we can hook up a bundle size analyser to avoid further regressions

PR #51, opened 2023-08-04
Switch to graphemer to split strings

sandhose:
I had a problem with consuming the ESM module in vitest.
The resulting bundle had a `import { split } from "lodash"`, which does not work because lodash is not an ESM module.
There is an alternative `lodash-es` which is an ESM module, but then using that would break some CJS environments.

`lodash`'s split function notably doesn't handle well some emoji graphemes, so instead this switches to `graphemer`

fixes vector-im/compound#196

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,6 @@
     "eslint-plugin-react": "^7.33.1",
     "eslint-plugin-storybook": "^0.6.13",
     "jsdom": "^22.1.0",
-    "lodash-es": "^4.17.21",
     "prettier": "3.0.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "eslint-plugin-react": "^7.33.1",
     "eslint-plugin-storybook": "^0.6.13",
     "jsdom": "^22.1.0",
+    "lodash-es": "^4.17.21",
     "prettier": "3.0.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
@@ -98,8 +99,7 @@
   },
   "dependencies": {
     "@radix-ui/react-form": "^0.0.3",
-    "classnames": "^2.3.2",
-    "lodash": "^4.17.21"
+    "classnames": "^2.3.2"
   },
   "peerDependencies": {
     "@fontsource/inter": "^5",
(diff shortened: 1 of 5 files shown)


t3chguy:
image

Lodash currently makes up a not insignificant amount of the react SDK main bundle. This sounds like it'll double that.

src/utils/string.ts
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { split } from "lodash";
+import { split } from "lodash-es";
```
t3chguy, line 17: Also worth pointing out EW hit bugs in Lodash's split grapheme splitting, we had to switch to `graphemer`

t3chguy: See https://github.com/matrix-org/matrix-react-sdk/pull/10976

t3chguy: https://github.com/beaugunderson/emoji-aware/issues/25 & https://github.com/lodash/lodash/issues/5652 being related issues about this bug

sandhose:
Not that much, because this tree-shakes correctly, so only the methods we use (in this case, `_.split`) are included in the bundle.

Library size with lodash-es not marked as external:

```
dist/style.css        20.06 kB │ gzip: 3.39 kB
dist/compound-web.js  21.02 kB │ gzip: 6.36 kB
dist/compound-web.cjs  16.59 kB │ gzip: 5.71 kB
```

and with it marked as external:

```
dist/style.css        20.06 kB │ gzip: 3.39 kB
dist/compound-web.js  16.40 kB │ gzip: 4.68 kB
dist/compound-web.cjs  13.03 kB │ gzip: 4.19 kB
```

So there is a 1.5kB difference gzipped

src/utils/string.ts
t3chguy: Have opened https://github.com/vector-im/compound/issues/196 to track this

t3chguy:
> So there is a 1.5kB difference gzipped

Otherwise known as 36%...

sandhose:
Alright fair, switching to `graphemer` then

package.json
```diff
@@ -83,6 +82,7 @@
     "eslint-plugin-react": "^7.33.1",
     "eslint-plugin-storybook": "^0.6.13",
     "jsdom": "^22.1.0",
+    "lodash-es": "^4.17.21",
```
t3chguy, line 85: This can probably be removed

t3chguy (approved):
:shipit: bonus points for the flag test

PR #55, opened 2023-08-09
Add NPM release CI workflow

germain-gg:
@Half-Shot This should give us tagged releases like in https://github.com/matrix-org/eslint-plugin-matrix-org/releases 

Would that be enough for what you're trying to achieve?

### DIFF
--- a/.github/workflows/npm_release.yml
+++ b/.github/workflows/npm_release.yml
@@ -1,0 +1,62 @@
+name: Release npm package
+on:
+  workflow_dispatch:
+    inputs:
+      version-bump:
+        description: The scale of the version bump required for semver compatibility
+        required: true
+        default: patch
+        type: choice
+        options:
+          - patch
+          - minor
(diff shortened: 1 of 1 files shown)


.github/workflows/npm_release.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: 👊 Bump version
+        run: |
+          yarn version --no-git-tag-version --${{ github.event.inputs.version-bump }}
+          git config --global user.name 'ElementRobot'
+          git config --global user.email 'releases@riot.im'
+          git commit -am "${{ github.event.inputs.version-bump }} version bump"
+          git push
+      - name: ⚙️ Build
+        run: |
+          yarn build
```
t3chguy, line 45: You don't need this as your build script is identical to your `prepare` script which is called by publishing.

t3chguy (approved):

PR #57, opened 2023-08-10
Improvements to the Avatar component

germain-gg:
For https://github.com/matrix-org/matrix-react-sdk/pull/11393 and https://github.com/vector-im/compound/issues/137

Makes the button polymorphic and allows it to be a `button` as well as a normal Avatar.
The image now can load lazily as well.

And we have an improved Storybook story.

### DIFF
--- a/src/components/Avatar/Avatar.module.css
+++ b/src/components/Avatar/Avatar.module.css
@@ -37,8 +37,7 @@ button.avatar {
   cursor: pointer;
 }
 
-button.avatar:disabled,
-button.avatar[disabled] {
+button.avatar:disabled {
   cursor: not-allowed;
 }
 
--- a/src/components/Avatar/Avatar.module.css
+++ b/src/components/Avatar/Avatar.module.css
@@ -27,6 +27,21 @@ limitations under the License.
   user-select: none;
 }
 
+button.avatar {
+  /**
+   * The avatar can be a button element, we need to reset its style
+   */
+  padding: 0;
+  border: 0;
+  appearance: none;
+  cursor: pointer;
+}
+
+button.avatar:disabled,
+button.avatar[disabled] {
+  cursor: not-allowed;
+}
+
 .avatar,
 .image {
   aspect-ratio: 1 / 1;
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -26,25 +26,25 @@ type AvatarProps = (
   | JSX.IntrinsicElements["span"]
 ) & {
   /**
-   * The avatar image URL
+   * The avatar image URL, if any.
    */
   src?: React.ComponentProps<typeof SuspenseImg>["src"];
   /**
    * The Matrix ID, Room ID, or Alias to generate the color when no image source
-   * is provided. Also used as a fallback when name is empty
+   * is provided. Also used as a fallback when name is empty.
    */
   id: string;
   /**
-   * The name used for the initial letter displayed when no image source is provided
+   * The name used for the initial letter displayed when no image source is provided.
    */
   name: string;
   /**
-   * Defines the avatar type, typically round, square is usually for spaces
+   * Defines the avatar type, typically round, square is usually for spaces.
    * @default "round"
    */
   type?: "square" | "round";
   /**
-   * The avatar size
+   * The avatar size in CSS units, e.g. `"24px"`.
    */
   size?: CSSStyleDeclaration["height"];
   /**
@@ -60,7 +60,7 @@ type AvatarProps = (
    */
   onKeyUp?: (e: React.KeyboardEvent) => void;
   /**
-   * Callback when the image has failed to load
+   * Callback when the image has failed to load.
    */
   onError?: React.ComponentProps<typeof SuspenseImg>["onError"];
 };
@@ -76,7 +76,7 @@ function shouldBeAButton(props: Partial<AvatarProps>): boolean {
 
 /**
  * Avatar component that will fallback to an initial letter over a coloured
- * background if no source is provided or if the source has failed to load
+ * background if no source is provided or if the source has failed to load.
  */
 export const Avatar = forwardRef<
   HTMLSpanElement | HTMLButtonElement,
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -25,12 +25,43 @@ type AvatarProps = (
   | JSX.IntrinsicElements["button"]
   | JSX.IntrinsicElements["span"]
 ) & {
+  /**
+   * The avatar image URL
+   */
   src?: React.ComponentProps<typeof SuspenseImg>["src"];
+  /**
+   * The Matrix ID, Room ID, or Alias to generate the color when no image source
+   * is provided. Also used as a fallback when name is empty
+   */
   id: string;
+  /**
+   * The name used for the initial letter displayed when no image source is provided
+   */
   name: string;
+  /**
+   * Defines the avatar type, typically round, square is usually for spaces
+   * @default "round"
+   */
   type?: "square" | "round";
+  /**
+   * The avatar size
+   */
   size?: CSSStyleDeclaration["height"];
+  /**
+   * On click handler, will turn the avatar into a button
+   */
   onClick?: (e: React.MouseEvent) => void;
+  /**
+   * Key down handler, will turn the avatar into a button
+   */
+  onKeyDown?: (e: React.KeyboardEvent) => void;
+  /**
+   * Key up handler, will turn the avatar into a button
+   */
+  onKeyUp?: (e: React.KeyboardEvent) => void;
+  /**
+   * Callback when the image has failed to load
+   */
   onError?: React.ComponentProps<typeof SuspenseImg>["onError"];
 };
 
@@ -43,6 +74,10 @@ function shouldBeAButton(props: Partial<AvatarProps>): boolean {
   return !!(props.onClick || props.onKeyDown || props.onKeyUp);
 }
 
+/**
+ * Avatar component that will fallback to an initial letter over a coloured
+ * background if no source is provided or if the source has failed to load
+ */
 export const Avatar = forwardRef<
   HTMLSpanElement | HTMLButtonElement,
   AvatarProps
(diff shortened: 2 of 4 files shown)


src/components/Avatar/Avatar.module.css
```diff
@@ -27,6 +27,21 @@ limitations under the License.
...
+  /**
+   * The avatar can be a button element, we need to reset its style
+   */
+  padding: 0;
+  border: 0;
+  appearance: none;
+  cursor: pointer;
+}
+
+button.avatar:disabled,
```
weeman1337, line 40: Should the pseudo class not be enough here?

weeman1337 (approved):
LGTM. Tested the stories :+1:

src/components/Avatar/Avatar.tsx
```diff
@@ -21,16 +21,24 @@ import { SuspenseImg } from "../../utils/SuspenseImg";
...
   name: string;
   type?: "square" | "round";
   size?: CSSStyleDeclaration["height"];
+  onClick: (e: React.MouseEvent) => void;
   onError?: React.ComponentProps<typeof SuspenseImg>["onError"];
 };
 
-export const Avatar = forwardRef<HTMLSpanElement, AvatarProps>(function Avatar(
+export const Avatar = forwardRef<
+  HTMLSpanElement | HTMLButtonElement,
```
t3chguy, line 39: Could this please be generic depending on `onClick`'s presence or `as` so we don't need to do type assertions on the caller side

src/components/Avatar/Avatar.tsx
```diff
@@ -21,16 +21,24 @@ import { SuspenseImg } from "../../utils/SuspenseImg";
...
+  | JSX.IntrinsicElements["button"]
+  | JSX.IntrinsicElements["span"]
+) & {
+  as: "button" | "span";
   src?: React.ComponentProps<typeof SuspenseImg>["src"];
   id: string;
   name: string;
   type?: "square" | "round";
   size?: CSSStyleDeclaration["height"];
+  onClick: (e: React.MouseEvent) => void;
```
t3chguy, line 34: Can we please protect against span+onClick at the type level, this is poor for a11y

germain-gg: That was my original intent, to use type unions, but the combination of everything i tried resulted with TS being very unhappy.

t3chguy: If you cannot do it as such then maybe the combination is bad and you should be more declarative and need to wrap it in a semantic button component as the caller?

src/components/Avatar/Avatar.tsx
t3chguy: Yeah this gets a lot trickier with forwardRef but I think bad types here will yield landmines and footguns for callers who are unfamiliar with compound.

nadonomy (approved):
The new storybook story is super useful, thx!

src/components/Avatar/Avatar.tsx
germain-gg: I have removed the `as`, and infer the HTML tag from the `onClick` prop. 
If we're using `onClick` then the avatar switches to be a `button`

src/components/Avatar/Avatar.tsx
germain-gg: I think this expresses the reality of the ref, it can either be a button or a span. I can make it more generic and say `HTMLElement` if you think that brings more value 🤷

src/components/Avatar/Avatar.tsx
```diff
@@ -40,46 +47,52 @@ export const Avatar = forwardRef<HTMLSpanElement, AvatarProps>(function Avatar(
...
-            height={size}
-            title={id}
-            onError={onError}
-          />
-        </Suspense>
-      )}
-    </span>
+  return React.createElement(
+    // Preventing `span` elements to have an `onClick` as it is usually
+    // terrible for accesibility
```
t3chguy, line 60: Always* as keyboard-only users can't click on things, so you'd need to also handle onKeyDown & onKeyUp

src/components/Avatar/Avatar.tsx
t3chguy: Can you document it at least `onClick ? button : span` but in English? Is the intention for the storybook to be the developer documentation too? Compound seems to be lacking jsdoc or any documentation for what props mean. Like type=square doesn't say anything about it being a rounded square, onClick doesn't say it'll change the underlying element etc. The overall component doesn't say what its for, that it generates a fallback avatar. I think the lack of documentation puts us in a far worse place than using BaseAvatar in react-sdk

t3chguy (changes requested):
Needs better developer-facing documentation for props & overall component usage & behaviours

src/components/Avatar/Avatar.tsx
```diff
@@ -25,12 +25,43 @@ type AvatarProps = (
   | JSX.IntrinsicElements["button"]
   | JSX.IntrinsicElements["span"]
 ) & {
+  /**
+   * The avatar image URL
```
t3chguy, line 29:
```suggestion
   * The avatar image URL, if any.
```

src/components/Avatar/Avatar.tsx
```diff
@@ -25,12 +25,43 @@ type AvatarProps = (
...
+   */
   src?: React.ComponentProps<typeof SuspenseImg>["src"];
+  /**
+   * The Matrix ID, Room ID, or Alias to generate the color when no image source
+   * is provided. Also used as a fallback when name is empty
```
t3chguy, line 34:
```suggestion
   * is provided. Also used as a fallback when name is empty.
```

src/components/Avatar/Avatar.tsx
```diff
@@ -25,12 +25,43 @@ type AvatarProps = (
...
+   */
   id: string;
+  /**
+   * The name used for the initial letter displayed when no image source is provided
```
t3chguy, line 38:
```suggestion
   * The name used for the initial letter displayed when no image source is provided.
```

src/components/Avatar/Avatar.tsx
```diff
@@ -25,12 +25,43 @@ type AvatarProps = (
...
+   */
   name: string;
+  /**
+   * Defines the avatar type, typically round, square is usually for spaces
```
t3chguy, line 42:
```suggestion
   * Defines the avatar type, typically round, square is usually for spaces.
```

src/components/Avatar/Avatar.tsx
germain-gg: Have added JSDoc comments, it was definitely lacking, and I will actually document all the other components with JSDoc comment to make sure this practise spreads across the codebase. Thank you for calling that out.

We would likely not ever document implementation details like "square has rounded corners". But as discussed out of band, documenting the underlying element used, can be greatly beneficial from an a11y point of view.

src/components/Avatar/Avatar.tsx
```diff
@@ -25,12 +25,43 @@ type AvatarProps = (
...
+   * @default "round"
+   */
   type?: "square" | "round";
+  /**
+   * The avatar size
```
t3chguy, line 47:
```suggestion
   * The avatar size in CSS units, e.g. `"24px"`.
```

src/components/Avatar/Avatar.tsx
```diff
@@ -25,12 +25,43 @@ type AvatarProps = (
...
+   */
   onClick?: (e: React.MouseEvent) => void;
+  /**
+   * Key down handler, will turn the avatar into a button
+   */
+  onKeyDown?: (e: React.KeyboardEvent) => void;
+  /**
+   * Key up handler, will turn the avatar into a button
+   */
+  onKeyUp?: (e: React.KeyboardEvent) => void;
```
t3chguy, line 61: Can these not be inferred from the button & span props?

src/components/Avatar/Avatar.tsx
```diff
@@ -25,12 +25,43 @@ type AvatarProps = (
...
+  /**
+   * Callback when the image has failed to load
+   */
```
t3chguy, line 64: Same as above

src/components/Avatar/Avatar.tsx
```diff
@@ -43,6 +74,10 @@ function shouldBeAButton(props: Partial<AvatarProps>): boolean {
   return !!(props.onClick || props.onKeyDown || props.onKeyUp);
 }
 
+/**
+ * Avatar component that will fallback to an initial letter over a coloured
+ * background if no source is provided or if the source has failed to load
```
t3chguy, line 79:
```suggestion
 * background if no source is provided or if the source has failed to load.
```

src/components/Avatar/Avatar.tsx
germain-gg: They can, they are only here for the JSDoc.

src/components/Avatar/Avatar.tsx
```diff
@@ -21,16 +21,67 @@ import { SuspenseImg } from "../../utils/SuspenseImg";
...
+  /**
+   * Key down handler, will turn the avatar into a button
+   */
+  onKeyDown?: (e: React.KeyboardEvent) => void;
+  /**
+   * Key up handler, will turn the avatar into a button
+   */
+  onKeyUp?: (e: React.KeyboardEvent) => void;
+  /**
+   * Callback when the image has failed to load
```
t3chguy, line 63:
```suggestion
   * Callback when the image has failed to load.
```

src/components/Avatar/Avatar.tsx
```diff
@@ -21,16 +21,67 @@ import { SuspenseImg } from "../../utils/SuspenseImg";
...
+   */
+  onClick?: (e: React.MouseEvent) => void;
+  /**
+   * Key down handler, will turn the avatar into a button
+   */
+  onKeyDown?: (e: React.KeyboardEvent) => void;
+  /**
+   * Key up handler, will turn the avatar into a button
+   */
+  onKeyUp?: (e: React.KeyboardEvent) => void;
```
t3chguy, line 61:
```suggestion
  /**
   * On click handler, will turn the avatar into a button element.
   */
  onClick?: (e: React.MouseEvent) => void;
  /**
   * Key down handler, will turn the avatar into a button element.
   */
  onKeyDown?: (e: React.KeyboardEvent) => void;
  /**
   * Key up handler, will turn the avatar into a button element.
   */
  onKeyUp?: (e: React.KeyboardEvent) => void;
```

t3chguy (approved):
Ideally we'd use techniques from https://stackoverflow.com/a/58473012 to fix the Ref type to match the underlying html element in use

PR #58, opened 2023-08-10
Add interoperability for graphemer import

germain-gg:
To work around https://github.com/flmnt/graphemer/issues/11

This will fix the failing jest tests in https://github.com/matrix-org/matrix-react-sdk/pull/11393
Runtime works because it is using the ESM module.

Thank you @sandhose for the pointer 🙇


sandhose (approved):
js packaging is fun

SimonBrandner (approved):

PR #60, opened 2023-08-11
Remove React.Suspense from Avatar

germain-gg:
It creates a lot more problems than it solves on the testing side of things on downstream projects.
We should instead strive to provide UI elements and leave individual products to make their fallback strategies


src/utils/SuspenseImg.tsx
germain-gg, line 1: Keeping as this will be used, but not for this component yet.

SimonBrandner (approved):

PR #63, opened 2023-08-23
Create Search component

germain-gg:
Screenshot 2023-08-23 at 13 57 41

For https://github.com/vector-im/compound/issues/213
Figma https://www.figma.com/file/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?type=design&node-id=797-8218&mode=design&t=ZthOhaEMDShBQiN4-11


src/components/Search/Search.tsx
```diff
@@ -0,0 +1,73 @@
...
+  className?: string;
+  /**
+   * The input placeholder.
+   * @default Search
+   */
+  placeholder?: string;
+  /**
+   * The field name.
+   */
+  name: React.ComponentProps<typeof Field>["name"];
```
weeman1337, line 37: This prop is not used. Should it be set as field name?

src/components/Search/Search.tsx
```diff
@@ -0,0 +1,73 @@
...
+import styles from "./Search.module.css";
+import { Field, Label } from "../Form";
+
+import SearchIcon from "@vector-im/compound-design-tokens/icons/search.svg";
+
+type SearchProps = {
+  /**
+   * The CSS class name
+   */
+  className?: string;
```
weeman1337, line 28: This is the class name of the `Label` wrapped by `Field`. From a developer perspective it would be good to be able to set the container class as well.

What about adding this to the `Field` classes instead? Then you can tweak everything, if required.

src/components/Search/Search.module.css
```diff
@@ -0,0 +1,64 @@
...
+  border: 1px solid var(--cpd-color-border-interactive-secondary);
+  border-radius: 9999px;
+  height: 36px;
+  box-sizing: border-box;
+  color: var(--cpd-color-text-primary);
+  display: inline-flex;
+  gap: var(--cpd-space-2x);
+  align-items: center;
+  padding: var(--cpd-space-1-5x) var(--cpd-space-3x);
+  overflow: hidden;
```
weeman1337, line 27: This shouldn't be necessary because there is an input field that holds the variable text.

src/components/Search/Search.module.css
```diff
@@ -0,0 +1,64 @@
...
+limitations under the License.
+*/
+
+.search {
+  border: 1px solid var(--cpd-color-border-interactive-secondary);
+  border-radius: 9999px;
+  height: 36px;
+  box-sizing: border-box;
+  color: var(--cpd-color-text-primary);
+  display: inline-flex;
```
weeman1337, line 23: The `.search` element is wrapped by a `Field`. I would just set `flex` here, so that the wrapper controls the actual size.

weeman1337 (changes requested):
Looks good. Some notes:

- The disabled state is not really discoverable (It looks the same than enabled). But it is also not in design. Is this going to change?
- The icon is too small compared to the designs

image

germain-gg:
Back in draft for now as https://github.com/matrix-org/matrix-react-sdk/pull/11507 will alleviate the concern and need for that component in the short term

src/components/Search/Search.tsx
germain-gg: Using `asChild` instead so the field is merged with the label. That solves the extra DOM element issue.

src/components/Search/Search.module.css
germain-gg: Not anymore, i have set `asChild` on the `Field` component

germain-gg:
The is no disabled state for now, henced removed it from the props the component receives. Might be added at a later stage.

weeman1337 (changes requested):
Sizes are good now :+1: 

The placeholder colours still needs a tweak: Design uses the placeholder colour where the component uses text primary.

weeman1337:
Also dark theme doesn't look so good

image

germain-gg:
The placeholder is set correctly, you can validate it by checking the shadow dom element as explained on https://stackoverflow.com/questions/26852922/inspect-webkit-input-placeholder-with-developer-tools

weeman1337 (approved):
Looks good :+1:

PR #68, opened 2023-09-04
Make sure all the CSS variables used exist (& upgrade the design tokens)

sandhose:
After #67, I saw a bunch of places which were using undefined CSS variables. This adds a stylelint plugin to make sure only defined variables are being used.

While looking at this, I saw a bunch of issues that I fixed along the way:

 - the Typography CSS module was using attr-based matches on the class attribute, which did not work with the CSS modules classname hashing
 - the IconButton had the wrong height in Safari, so I fixed it
 - `stylelint` was actually not being run because of some shell quoting issue
 - the latest versions of the design tokens broke some stuff (because of icons being renamed) so I upgraded them


.stylelintrc.json
```diff
@@ -1,4 +1,14 @@
 {
   "extends": "stylelint-config-standard",
+  "plugins": ["stylelint-value-no-unknown-custom-properties"],
+  "rules": {
+    "csstools/value-no-unknown-custom-properties": [true, {
+      "importFrom": [
+        "./src/styles/global.css",
+        "./node_modules/@vector-im/compound-design-tokens/assets/web/css/cpd-common.css",
+        "./node_modules/@vector-im/compound-design-tokens/assets/web/css/cpd-light-mq.css"
```
sandhose, line 9: I chose one of the themes to define the variables

src/components/Button/Button.module.css
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 .button {
   border-radius: var(--cpd-radius-pill-effect);
   cursor: pointer;
-  -webkit-appeareance: none;
```
sandhose, line 20: That was a typo not being caught by stylelint because of the shell expansion issue

src/components/Form/Controls/Password/Password.tsx
```diff
@@ -18,19 +18,19 @@ import React, { forwardRef, PropsWithChildren, useReducer } from "react";
 import { ActionControl } from "../../../ActionControl/ActionControl";
 import { Control } from "../../Control";
 
-import VisibilityVisible from "@vector-im/compound-design-tokens/icons/visibility-visible.svg";
-import VisibilityInvisible from "@vector-im/compound-design-tokens/icons/visibility-invisible.svg";
+import VisibilityOn from "@vector-im/compound-design-tokens/icons/visibility-on.svg";
+import VisibilityOff from "@vector-im/compound-design-tokens/icons/visibility-off.svg";
```
sandhose, line 22: Those icons were renamed in the design tokens

src/components/Typography/Typography.module.css
```diff
@@ -18,158 +18,135 @@ limitations under the License.
  * Font – Body – Extra Small
  */
 
-[class^="font-body-xs"] {
```
sandhose, line 21: Looking at the CSS output file makes it obvious that this trick does not work with CSS modules

src/styles/global.css
```diff
@@ -17,6 +17,10 @@ limitations under the License.
 
 :root {
   --cpd-radius-pill-effect: 9999px;
+
+  /* Default icon and avatar size */
+  --cpd-icon-button-size: var(--cpd-space-8x);
+  --cpd-avatar-size: var(--cpd-space-16x);
```
sandhose, line 23: This is mainly to avoid the "variable is not defined" error for those, even though they are always set within the component

robintown (commented):
One quick question

src/components/Typography/Typography.module.css
```diff
@@ -18,158 +18,135 @@ limitations under the License.
...
   font: var(--cpd-font-heading-md-regular);
 }
 
 .font-heading-md-semibold {
+  letter-spacing: var(--cpd-font-letter-spacing-heading-md);
   font: var(--cpd-font-heading-md-semibold);
 }
 
-.font-heading-md-medium {
-  font: var(--cpd-font-heading-md-medium);
```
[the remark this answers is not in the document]
sandhose: They indeed don't exist in the design tokens. For the `heading` variant, there is only the `regular` and `semibold` weights defined You can see all the defined ones here:

https://github.com/vector-im/compound-design-tokens/blob/75c35f1ce192e6a4a604226090d67f1eeaecd91a/assets/web/css/cpd-common.css#L92-L112

robintown (approved):

sandhose:
When I merged back `main`, turns out the recent changes on the `Button` component (#72) were not passing stylelint, so I changed the new class to be `kebab-case` instead of `snakeCase`

PR #70, opened 2023-09-07
Document a gotcha where linking this package leads to conflicting @types/react copies

robintown:
I spent a good couple hours today struggling to `yarn link` this package in a way that wouldn't create type errors in my Element Call checkout, and feel compelled to write down my workaround somewhere that others can find it.

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -29,7 +29,7 @@ const config: StorybookConfig = {
 
       publicDir: "res",
     };
-  }
+  },
 };
 
 export default config;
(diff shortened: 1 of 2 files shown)


andybalaam (approved):
Looks great, thanks!

robintown:
I did try that, yeah, but even with identical versions in node\_modules/@types/react and node\_modules/@vector-im/compound-web/node\_modules/@types/react, I was getting the same set of 182 errors… not sure what to make of that since it somehow works for Element Web

PR #74, opened 2023-09-08
Fix a regression in the button alignment

sandhose:
Because of #72, the inner text of the button became left-aligned.
Before, it was center-aligned, which was important if the button was bigger than its natural size.
This fixes that.

Before:
image

After:
image

No changes when the button size is its natural one:
image

For the record, this affects MAS' pagination buttons:

image


weeman1337 (approved):
Thank you @sandhose . LGTM :+1:

PR #77, opened 2023-09-15
Fix the AvatarStack component tests

sandhose:
The tests were just wrong. The test files were not typechecking correctly, which apparently we don't do as part of linting?


t3chguy (approved):
LGTM 

> The test files were not typechecking correctly, which apparently we don't do as part of linting?

is concerning, can you open an issue?

sandhose:
> > The test files were not typechecking correctly, which apparently we don't do as part of linting?
> 
> 
> 
> is concerning, can you open an issue?

Yes please

PR #78, opened 2023-09-18
Run `tsc` as part of the lint script and upgrade dependencies

sandhose:


src/global.d.ts
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2023 New Vector Ltd.
+Copyright 2023  New Vector Ltd
```
t3chguy, line 2: This looks wrong?

tsconfig.json
```diff
@@ -1,33 +1,11 @@
...
-    "typeRoots": ["./node_modules/@types", "./src/@types"],
-    "types": ["node"],
-
-    /* Completeness */
-    "skipLibCheck": true /* Skip type checking all .d.ts files. */
-  }
+    "types": ["vite/client"]
+  },
+  "include": ["src"],
+  "exclude": ["**/*.stories.tsx"],
```
t3chguy, line 9: Does this not mean errors like the one in your recent PR will continue to be permissable?

src/global.d.ts
sandhose: Woops, I actually re-created that file and copy-pasted the copyright from the wrong place. Fixed

tsconfig.json
sandhose: Hah, this was me trying to figure out why the `vite-plugin-dts` was not working as expected and I forgot to remove it. Fixed

.storybook/main.ts
```diff
@@ -7,7 +7,7 @@ const config: StorybookConfig = {
     "@storybook/addon-essentials",
     "@storybook/addon-interactions",
     "@storybook/addon-designs",
-    "@storybook/addon-styling",
```
sandhose, line 10: `addon-styling` is deprecated

.storybook/main.ts
```diff
@@ -22,11 +22,6 @@ const config: StorybookConfig = {
   viteFinal: (config) => {
     return {
       ...config,
-
-      // In the `vite.config.ts` file, we set the `experimental.renderBuiltUrl`
-      // option, which breaks the storybook build. This clears that option.
-      experimental: undefined,
```
sandhose, line 28: That was an unnecessary leftover

package.json
```diff
@@ -50,58 +51,58 @@
...
     "vite-plugin-svgr": "^3.2.0",
-    "vitest": "^0.33.0"
+    "vitest": "^0.34.4"
   },
   "dependencies": {
     "@radix-ui/react-form": "^0.0.3",
     "@radix-ui/react-tooltip": "^1.0.6",
     "classnames": "^2.3.2",
-    "graphemer": "^1.4.0",
-    "rimraf": "^3.0.1"
```
sandhose, line 104: No idea why it was in the package dependencies and not devDependencies

package.json
```diff
@@ -118,5 +119,8 @@
     "@types/react-dom": {
       "optional": true
     }
+  },
+  "resolutions": {
+    "jackspeak": "2.1.1"
```
sandhose, line 124: Workaround for https://github.com/storybookjs/storybook/issues/22431#issuecomment-1630086092

PR #80, opened 2023-09-18
Disable hover styles on touchscreen devices

robintown:
Browsers on devices like phones and tablets often use an invisible, simulated cursor as a way to let mobile users activate the hover states of elements. This means that when you press a Compound button, on mobile, the button will stay in the hover state until you tap somewhere else on the page. This looks unnatural, so I suggest we opt out of this behavior for devices whose primary input device does not support hovering.

We can, however, refer to the :active pseudo-class to still display these hover states for the duration that the element is being pressed.

### DIFF
--- a/src/components/Alert/Alert.module.css
+++ b/src/components/Alert/Alert.module.css
@@ -72,8 +72,15 @@ limitations under the License.
   cursor: pointer;
 }
 
-.close:hover {
+@media (hover) {
+  .close:hover {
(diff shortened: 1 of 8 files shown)


weeman1337 (approved):
Looks good :+1:

robintown:
Right, I don't have bandwidth to do that at the moment, but it sounds very valuable, since writing hover styles outside of a `@media (hover)` block is very easy to forget. I'll put it on my to-do list!

PR #85, opened 2023-09-25
Ponyfill useId

Johennes:
Inspired by https://github.com/ezcater/recipe/blob/dc253df74f0b9c509f13172980eb8c1a0defb62a/packages/recipe/src/utils/hooks/useUniqueId.ts.

Fixes: vector-im/compound#224

### DIFF
--- a/src/components/ActionControl/ActionControl.tsx
+++ b/src/components/ActionControl/ActionControl.tsx
@@ -15,10 +15,11 @@ limitations under the License.
 */
 
 import classnames from "classnames";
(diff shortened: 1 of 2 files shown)


Johennes:
I seem to have upset Percy and Sonarcloud but I can't quite see how 🤔

t3chguy:
image

@Johennes this repo doesn't correctly work with forks

Johennes:
Oh, I see. Seems worth fixing in the long run but I've re-pushed this as https://github.com/vector-im/compound-web/pull/88 now to unblock.

PR #88, opened 2023-09-26
Ponyfill useId

Johennes:
Inspired by https://github.com/ezcater/recipe/blob/dc253df74f0b9c509f13172980eb8c1a0defb62a/packages/recipe/src/utils/hooks/useUniqueId.ts.

Fixes: vector-im/compound#224
Replaces: https://github.com/vector-im/compound-web/pull/85

### DIFF
--- a/src/utils/useId.ts
+++ b/src/utils/useId.ts
@@ -1,3 +1,19 @@
+/*
+Copyright 2023 New Vector Ltd
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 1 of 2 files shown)


Johennes:
Percy is now failing because of the quota.

> [percy] Error: This organization has exceeded the limits of the Percy BrowserStack plan. Administrators can upgrade here: https://percy.io/organizations/c8fecada/billing

src/utils/useId.ts
```diff
@@ -0,0 +1,16 @@
+import * as React from "react";
```
t3chguy, line 1: copyright pls

t3chguy (approved):

src/utils/useId.ts
Johennes: Ah geez. You can tell I don't get to add new files very often. 🤦‍♂️

Thanks, should be fixed now.

germain-gg (commented):
Code looks good to me,

Could we add an ESLint rule to make `useId` a restricted import and suggest using your ponyfill instead? 
I believe we're using https://eslint.org/docs/latest/rules/no-restricted-imports on EW

Does not have to block this pull request, but a quick follow up would be appreciated

Johennes:
Ok, let me see if I can get that going tonight. Will land this in the interim.

Johennes:
Sorry, I didn't get around to do this last night. I have put in https://github.com/vector-im/compound/issues/228 and will try to come back to it.

PR #89, opened 2023-09-26
Check formatting in CI

robintown:


robintown:
Replaced by https://github.com/vector-im/compound-web/pull/89 since CI is broken for forks

PR #91, opened 2023-09-26
Create menu item component

robintown:
[Latest designs](https://www.figma.com/file/fq7cTHPNHZX3ttFW1gBEnv/Compound-Element-Call-Components?type=design&node-id=56-41978&mode=dev)

Screenshot 2023-09-27 at 13-53-37 MenuItem - Primary ⋅ Storybook


src/components/MenuItem/MenuItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+   * The label to show on this menu item.
+   */
+  label: string;
+  /**
+   * The color variant of the menu item.
+   * @default primary
+   */
+  kind?: "primary" | "critical";
+  /**
+   * Whether to replace the children with a navigation hint on hover.
```
germain-gg, line 51: This comment does not look accurate compared to the implementation.
It seems to add a chevron rather than it replacing the children?

src/components/MenuItem/MenuItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+type MenuItemElement = "button" | "label" | "div";
+
+type Props<C extends MenuItemElement> = {
+  /**
+   * The element type of this menu item.
+   * @default button
+   */
+  as?: C;
+  className?: string;
```
germain-gg, line 36: Missing a JSDoc comment here

src/components/MenuItem/MenuItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+   */
+  kind?: "primary" | "critical";
+  /**
+   * Whether to replace the children with a navigation hint on hover.
+   * @default true
+   */
+  navHint?: boolean;
+} & ComponentPropsWithoutRef<C>;
+
+export const MenuItem = <C extends MenuItemElement = "button">({
```
germain-gg, line 57: Can we add a JSDoc comment here please

src/components/MenuItem/MenuItem.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export const Critical = Template.bind({});
+Critical.args = { kind: "critical" };
+
+Primary.parameters = Critical.parameters = {
+  design: {
+    type: "figma",
+    url: "https://www.figma.com/file/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?type=design&node-id=712-6909&mode=dev",
+  },
+};
```
germain-gg, line 54: Actually this is not needed anymore, we need to remove that from the `gen:component` script

That plugin does not work with Storybook 7 AFAIK

src/components/MenuItem/MenuItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * The color variant of the menu item.
+   * @default primary
+   */
+  kind?: "primary" | "critical";
+  /**
+   * Whether to replace the children with a navigation hint on hover.
+   * @default true
+   */
+  navHint?: boolean;
```
germain-gg, line 54: I'm unsure about that naming. It appears fine, but I don't understand what conditions warrant the chevron to be hidden.
I've raised that question in an internal room, and hopefully @americanrefugee might be able to add extra context and we can settle on something final here.

germain-gg (changes requested):
Looks like a strong start! Added a few comments, but overall seems on the right track

src/components/MenuItem/MenuItem.tsx
robintown: The use case is when you have `<input>`s as the child, for example in the [toggle menu item](https://github.com/vector-im/compound-web/pull/92) scenario. These shouldn't be hidden on hover, because that would make it awkward to interact with them, and the purpose of the menu item isn't for navigation.

src/components/MenuItem/MenuItem.tsx
germain-gg: The chevron needs to be displayed all the time. it's a mistake on the Figma design that it is missing.
We can therefore remove that prop

src/components/MenuItem/MenuItem.tsx
germain-gg: right, so i believe we could infer the visibility of the chevron based on whether the component has an `href` or `onClick` handler. no?

robintown: sure, makes sense to me

src/components/MenuItem/MenuItem.tsx
robintown: I guess we figured this out in [in discussion](https://github.com/vector-im/compound-web/pull/91#discussion_r1338556370)

robintown: oh, but: the replacement of the children happens in CSS, and I can see why that might not be obvious. I've added a couple comments to hopefully make that clearer.

src/components/MenuItem/MenuItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+  ComponentPropsWithoutRef,
+  ComponentType,
+  ElementType,
+  SVGAttributes,
+} from "react";
+import styles from "./MenuItem.module.css";
+import { Text } from "../Typography/Text";
+import ChevronRightIcon from "@vector-im/compound-design-tokens/icons/chevron-right.svg";
+
+type MenuItemElement = "button" | "label" | "div";
```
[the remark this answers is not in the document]
robintown: done

germain-gg (approved):
Looks great thank you very much for the updates!

PR #92, opened 2023-09-26
Create toggle menu item component

robintown:
:warning: **Based on https://github.com/vector-im/compound-web/pull/91**, please only review the new commits

This is a pretty common way for the menu item component to be used, and it's slightly non-trivial (you need to hook up the entire menu item as a label element), so I think it deserves its own component.

Screenshot 2023-09-26 at 15-03-46 ToggleMenuItem - Primary ⋅ Storybook


src/components/Toggle/Toggle.tsx
```diff
@@ -35,21 +31,11 @@ type ToggleProps = {
...
-      <input
-        ref={ref}
-        {...props}
-        type="checkbox"
-        onMouseDown={(e) => {
-          // Prevents the `focus` event from being fired when clicked
-          // but still reliably works when navigating via the keyboard
-          e.preventDefault();
-          onMouseDown?.(e);
-        }}
```
robintown, line 51: Removed, because I'm pretty sure this was a hack, for which a better solution is the `:focus` → `:focus-visible` change I made above. It wasn't preventing the focus ring from being shown when you click on a toggle menu item.

t3chguy: We should really be using something like https://www.npmjs.com/package/what-input or https://www.npmjs.com/package/focus-visible - some accessible technologies sometimes have to simulate clicks, the focus should remain on the item when that click occurs rather than be prevented. We should be using native elements where possible and *MUST* mimic their behaviour otherwise.

germain-gg: I'd be interested in reading more about those assistive technologies who do clicks like that. Want to understand that problem better and see how to circumvent it ?

germain-gg: Reading through other articles, https://hidde.blog/focus-visible-more-than-keyboard/ 
It does seems that the `:focus-visible` already applies styles to more non-mouse cases than just keyboard users.

In any case this should not block this pull request, and probably warrants a ticket to investigate and prove that we need to solve this issue.

t3chguy: Yeah definitely not a blocker here, @robintown's change is a move in the correct direction

germain-gg (approved):
Code looks good to me here!

PR #97, opened 2023-09-29
Create glass component

robintown:
There isn't necessarily a single place in the Compound Figma files where this glass effect has been componentized, but in the interest of keeping it consistent, I suggest we implement it here in this project.

Screenshot from 2023-09-29 15-04-37
Screenshot from 2023-09-29 15-04-58


germain-gg (approved):
Code wise looks pretty sane to me!

I have created an epic https://github.com/vector-im/compound/issues/230 and almost requested a review with @janogarcia to get a sense of whether we want to adjust values.

robintown:
Sounds good 👍

janogarcia (commented):
@robintown Thanks for the screenshots! 

On the glass material, there's a few things we haven't fully defined yet on the design side.

Yesterday I explored how to [improve it for Light mode](https://www.figma.com/file/fq7cTHPNHZX3ttFW1gBEnv/Compound-Element-Call-Components?type=design&node-id=1014-22762&mode=design&t=k55MNjwygyzmdsDt-0), which might also affect the Dark mode. I'll be discussing those with Aaron today, so that we can deliver a more solid spec.

As for the current implementation, the only change I would suggest is changing the calculation of the outer border radius:

It's currently set to `border-radius: calc(var(--cpd-space-9x) - var(--cpd-border-width-1));` but we shouldn't subtract the border width here.

robintown:
OK, feel free to ping me with whatever spec updates that come up.

The border width calculation is deliberate, actually: the Figma specs call for the border to be an "inner border", but no such thing exists in CSS. A `border` is always created on the outside of the padding, and `border-radius` refers to the inner radius, rather than the outer radius, so we have to adjust the value accordingly. There is a less confusing way to achieve an inner border by using the `outline` property though, so I've gone ahead and replaced the code with that approach.

janogarcia:
Thanks, Robin!

It's actually not super relevant how borders are setup for Figma, as long as we get the desired output on the target platform (Web in this case).

I guess we could just change the sizing to `box-sizing: border-box` and get the inner border, but I haven't tested it against other constraints we may have in place.

But the point is not about how to implement the border, but making sure we set the expected corner radius value. 

For example, if we set `border-radius` to 4px, it will remain 4px, no matter if the outer border width is 1px, 10px, or 100px. That's why I meant we shouldn't subtract the outer border width when setting the border radius. We should set it at the expected final value.

It would have been different if the outer border was implemented with a `box-shadow`, as the rounding would get actually larger as the shadow spread value increases.

janogarcia:
Just realized that the PR has been already merged, but I'd still suggest addressing the outer corner radius as described in my previous comment.

janogarcia:
And now I realize you actually set the border radius as requested as part of the latest changeset. 😊👍

robintown:
Huh, I just realized I was wrong anyways: `border-radius` does appear to refer to the outer radius, rather than the inner one. But that doesn't change anything now that I'm using `outline`.

PR #98, opened 2023-10-02
Create floating menu component

robintown:
For https://github.com/vector-im/compound/issues/231

As an intermediate step towards implementing menus, this adds a component for the variant of menus used on desktop. To make it ready for external use, I still need to implement the drawer menu variant used on mobile, and then package it up in a more high-level component that also handles menu triggers.

Screenshot 2023-10-02 at 00-26-44 floatingmenu--floating-menu

[Design that I referenced](https://www.figma.com/file/fq7cTHPNHZX3ttFW1gBEnv/Compound-Element-Call-Components?type=design&node-id=39-14979&mode=dev)


t3chguy (changes requested):
@robintown could you please re-open this PR as a branch on this repo rather then your fork due to https://github.com/vector-im/compound/issues/80 preventing CI from running

src/components/FloatingMenu/FloatingMenu.tsx
```diff
@@ -0,0 +1,77 @@
...
+import useId from "../../utils/useId";
+import { Text } from "../Typography/Text";
+
+interface Props extends ComponentPropsWithoutRef<"div"> {
+  /**
+   * The size of the menu in px.
+   */
+  // TODO: Figure out whether this component should come with a standard set of
+  // T-shirt sizes
+  size: number;
```
germain-gg, line 29: Is that prop supposed to represent the `width` of the menu? If that's the case we should use `width` over `size` as the latter suggests we might have a more or less dense version of the menu

But more importantly, I don't think we should let consumers of this component define that value manually. A combination of `width: fit-content` and a `max-width` would be more appropriate in my opinion.

Probably worth to loop @americanrefugee and @janogarcia to chime in on the component behaviour

src/components/FloatingMenu/FloatingMenu.tsx
```diff
@@ -0,0 +1,77 @@
...
+   */
+  className?: string;
+  /**
+   * The menu contents.
+   */
+  children: ReactNode;
+}
+
+/**
+ * A floating menu (the kind of menu you would see on desktop).
```
germain-gg, line 45: Not entirely sure whether [`@private`](https://jsdoc.app/tags-private.html) in JSDoc would apply to a whole component?

src/components/FloatingMenu/FloatingMenu.tsx
```diff
@@ -0,0 +1,77 @@
...
+ */
+// This an internal component not intended for export! Consumers should use it
+// via the Menu or ContextMenu components (which at the time of writing, are not
+// yet built).
```
germain-gg, line 49: What's the difference between `Menu` and `ContextMenu`?

If this is an internal, could we instead bury this within the `ContextMenu` folder and not export it? So it can't be mistakenly imported and incorrectly used

src/components/FloatingMenu/FloatingMenu.module.css
```diff
@@ -0,0 +1,39 @@
...
+
+.menu {
+  border-radius: var(--cpd-space-3x);
+  background: var(--cpd-color-bg-canvas-default);
+
+  /* Here we use outline to create a border internal to the container */
+  outline: var(--cpd-border-width-1) solid
+    var(--cpd-color-border-interactive-secondary);
+  outline-offset: calc(-1 * var(--cpd-border-width-1));
+  box-shadow: 0 4px 24px 0 rgba(0 0 0 / 10%);
```
germain-gg, line 25: @janogarcia it looks like we're starting to use shadows like that in quite a few places. We were discussing making those tokens.
Should we pick this up again?

src/components/FloatingMenu/FloatingMenu.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+describe("Menu", () => {
+  it("renders", () => {
+    const { asFragment } = render(
+      <FloatingMenu size={200} title="Settings">
+        <MenuItem Icon={UserProfileIcon} label="Profile" />
+        <MenuItem Icon={NotificationsIcon} label="Notifications" />
+        <MenuItem Icon={ChatProblemIcon} label="Feedback" />
+        <MenuDivider />
+        <MenuItem kind="critical" Icon={LeaveIcon} label="Sign out" />
```
germain-gg, line 37: Could we add an example with a very long German word, so we can ensure the word wrapping works well with the menu item?

germain-gg (commented):
This looks like a great start!
This warrants being looked at by a few folks on the design team, but nothing major to report on

src/components/FloatingMenu/FloatingMenu.module.css
robintown: That would be fantastic. In Element Call I ended up creating some CSS variables for them, but it would be even nicer if they could be formalized as tokens belonging to Compound.

src/components/FloatingMenu/FloatingMenu.tsx
robintown: I don't think `@private` exists in TSDoc though, which as I understand it is the standard our doc comments follow?

src/components/FloatingMenu/FloatingMenu.tsx
robintown: For the first point: I've now expanded the comment

For the second point: I've renamed the folder to Menu to reflect the fact that it'll be buried as an implementation detail of the Menu component. Not sure what you mean about not exporting it though, because we need to export the component for Storybook usage and tests - as long as we don't export it from index.ts, that should be enough to make sure it stays internal, I think?

robintown:
Replaced by https://github.com/vector-im/compound-web/pull/99

PR #99, opened 2023-10-02
Create floating menu component

robintown:
For https://github.com/vector-im/compound/issues/231

As an intermediate step towards implementing menus, this adds a component for the variant of menus used on desktop. To make it ready for external use, I still need to implement the drawer menu variant used on mobile, and then package it up in a more high-level component that also handles menu triggers.

Screenshot 2023-10-17 at 23-29-10 FloatingMenu - Floating Menu ⋅ Storybook

[Design that I referenced](https://www.figma.com/file/fq7cTHPNHZX3ttFW1gBEnv/Compound-Element-Call-Components?type=design&node-id=39-14979&mode=dev)


robintown:
@janogarcia @americanrefugee Germain had a couple of questions for you, which you can find on the [previous pull request](https://github.com/vector-im/compound-web/pull/98) (I had to close it unfortunately since I messed up the branches)

src/components/MenuItem/MenuDivider.module.css
```diff
@@ -0,0 +1,23 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+.divider {
+  margin-inline: var(--cpd-space-4x);
```
germain-gg, line 18: I really like your use of logical properties! I think there's some lint rules we can add to force the project using that over the physical properties, and that would be a good step towards better RTL support.

I filed https://github.com/vector-im/compound/issues/233 for future reference

src/components/MenuItem/MenuItem.module.css
```diff
@@ -35,6 +41,11 @@ limitations under the License.
   grid-template: "icon ." auto / auto 1fr;
 }
 
+/* Items with interactive controls like toggles are larger by default */
+label.item {
+  min-inline-size: 246px;
```
germain-gg, line 46: I'm never a fan of those magic numbers. @americanrefugee or @janogarcia how did we land on that value for the menu width? Trying to come up with a formula that would convey a bit more context.

@robintown let's not block this PR on this detail though

germain-gg (approved):
This looks great! Thank you for the updates to the component.
Waiting on the designer reviews

src/components/MenuItem/MenuDivider.module.css
robintown: I agree that we should enforce this! PR here: https://github.com/vector-im/compound-web/pull/101

americanrefugee:
A few comments and answers to questions...

@germain-gg asked "What's the difference between Menu and ContextMenu?"
- There is no difference, I was just inconsistent with naming. I'm good with calling these Context Menu

@robintown asked "The main issue was, what makes this menu wider than the base width of 200px?"
- The design file is probably not 100% accurate. In principle,
   - Min width is 200px
   - Max width is 320px
   - Margin between text label and toggle should be 24px
   - Menu should grow horizontally to accommodate list items accordingly so that there are no word/line breaks up to 320px. If more room is needed, then break the line, but not the word. And if we run into a case where a single word makes the menu more than 320px, then that would be the only exception.

So, my only comment on the implementation depicted above is that the word "Benachrichtigungen" should fit on one line

@janogarcia do you disagree with any of the above?

robintown:
Finally getting around to merging this - @americanrefugee sounds good, thanks for the guidance!

src/components/MenuItem/MenuItem.module.css
robintown: Removed it

PR #100, opened 2023-10-03
Create menu component

robintown:
Design specs: [Figma document](https://www.figma.com/file/fq7cTHPNHZX3ttFW1gBEnv/Compound-Element-Call-Components?type=design&node-id=39-14979&mode=design&t=UdjwpVM14filmw2Z-11)

This component represents a menu that opens on a button press. Feedback on the API is very welcome - I'm pretty happy with how I managed to condense Radix's tree of menu components into a single component, but we need to make sure that it's intuitive for others too.

[Screencast from 2023-10-18 00-03-46.webm](https://github.com/vector-im/compound-web/assets/48614497/c694c6eb-1f71-4de2-a960-be235b3e7138)

[Screencast from 2023-10-18 00-04-52.webm](https://github.com/vector-im/compound-web/assets/48614497/18971e48-1b78-4a3b-9848-950ff9dca798)


src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * Event handler called when the open state of the menu changes.
+   */
+  onOpenChange: (open: boolean) => void;
+  /**
+   * The button that opens the menu. This must be a component that accepts a ref
+   * and spreads props.
+   * https://www.radix-ui.com/primitives/docs/guides/composition
+   */
+  trigger: ReactNode;
```
germain-gg, line 45: Those three props seems slightly contradictory? 
I imagine that `trigger` is the main way to open the menu and `open` is here more for tests and storybook?

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+interface Props {
+  /**
+   * The menu title.
+   */
+  title: string;
+  /**
+   * Whether the menu is open.
+   */
+  open: boolean;
```
germain-gg, line 35: We shouldn't have to pass that prop most of the time, as `trigger` is the main thing that will control the open state of the menu?

```suggestion
  open?: boolean;
```

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,118 @@
...
+  onOpenChange,
+  trigger,
+  children: childrenProp,
+  side = "bottom",
+  align = "center",
+}) => {
+  // Normally, the menu takes the form of a floating box. But on mobile (which
+  // we detect by looking for whether the primary input device is a
+  // touchscreen), the menu should morph into a drawer
+  const touchscreen = useTouchscreen();
```
germain-gg, line 79: This should use https://github.com/vector-im/compound-web/blob/main/src/utils/platform.ts#L23 instead

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const context = useMemo(
+    () => ({
+      component: touchscreen
+        ? ("Vaul drawer" as const)
+        : ("Radix dropdown menu" as const),
+      onOpenChange,
+    }),
+    [touchscreen, onOpenChange],
+  );
```
germain-gg, line 88: I'm not sure I understand this piece of code? A device is unlikely to morph from a touch device to a non touch device?

I would also suggest to remove references to `radix` or `vaul` from the context values, as this will not pass the test of time, and will become hard to grok the day we refactor to use another library.

src/components/Menu/MenuDivider.tsx
germain-gg, line 1: Definitely not to address as part of this pull request, but we do have dividers scattered across the app, and I believe we could generalise this component rather than specialising it. 

I doubt it carries some special styling

src/components/Menu/MenuItem.tsx
```diff
@@ -97,4 +132,17 @@ export const MenuItem = <C extends MenuItemElement = "button">({
       {children}
     </Component>
   );
+
+  if (interactive) {
+    switch (context?.component) {
+      case "Radix dropdown menu":
```
germain-gg, line 138: What happens if `interactive` is `true` and the component used is `vaul` ?
I don't quite follow this bit?

src/components/Menu/MenuItem.tsx
```diff
@@ -62,24 +74,47 @@ export const MenuItem = <C extends MenuItemElement = "button">({
...
   label,
+  onSelect,
   kind = "primary",
   children,
+  onClick: onClickProp,
   ...props
-}: Props<C>): JSX.Element => {
+}: Props<C>): ReactNode => {
   const Component = as ?? ("button" as ElementType);
+  const interactive = as !== "div";
```
germain-gg, line 84: Not sure we can

package.json
```diff
@@ -102,10 +102,12 @@
     "vitest": "^0.34.4"
   },
   "dependencies": {
+    "@radix-ui/react-dropdown-menu": "^2.0.6",
```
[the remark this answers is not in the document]
robintown: That's such a gotcha… I wonder if we could somehow make vite.config.ts automatically read from package.json. (That would be for another PR, though.)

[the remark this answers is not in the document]
germain-gg: Very much agree with you. I have created vector-im/compound-web#142 to alleviate this! That'd be quite important I agree

src/components/Menu/Menu.tsx
germain-gg: I think your reasoning is totally fine, let's go with that. We might need to do a similar exercise to what Quentin did with forms and formalise and appraoch to controlled vs uncontrolled components.

Being opiniated, hence promoting more consistency throughout the codebase

kerryarchibald: > when components are consistently designed to always be controlled rather than uncontrolled, this leads to more obvious code

In my experience this is true, and makes things more testable and predictable.

I created this issue to track this discussion https://github.com/vector-im/compound/issues/268

src/components/Menu/MenuItem.stories.tsx
```diff
@@ -24,7 +24,7 @@ import { MenuItem as MenuItemComponent } from "./MenuItem";
 import { Text } from "../Typography/Text";
 
 export default {
-  title: "MenuItem",
+  title: "Menu/MenuItem",
   component: MenuItemComponent,
   tags: ["autodocs"],
   argTypes: {},
```
kerryarchibald, line 30:
```suggestion
  argTypes: {},
  parameters: {
    design: {
      type: "figma",
      url: "https://www.figma.com/file/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?type=design&node-id=712-6909&mode=design&t=7zspBeYd4EBaDpzt-0",
    },
  },
```

src/components/Menu/MenuItem.module.css
```diff
@@ -41,11 +41,6 @@ limitations under the License.
   grid-template: "icon ." auto / auto 1fr;
 }
 
-/* Items with interactive controls like toggles are larger by default */
```
kerryarchibald, line 44: Can this use `Separator` instead now?

kerryarchibald (commented):
Hi! Sorry if this got lost a bit during the handover. Picking up G's review, it looks great to me. I've left a few nitpicky comments about documentation, and it would be great to swap out the `MenuDivider` for the generic `Separator`.

Thanks!

src/components/Menu/Menu.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+import LeaveIcon from "@vector-im/compound-design-tokens/icons/leave.svg";
+
+import { Menu as MenuComponent } from "./Menu";
+import { MenuItem } from "./MenuItem";
+import { MenuDivider } from "./MenuDivider";
+import { Button } from "../Button/Button";
+
+export default {
+  title: "Menu",
+  component: MenuComponent,
```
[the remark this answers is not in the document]
robintown: Okay, I spent a little bit of time trying to fix this but was unsuccessful given my lack of familiarity with Storybook, so I think I will leave this for another PR

src/components/Menu/MenuItem.stories.tsx
kerryarchibald: Hmm weird, they seem to work fine with latest storybook and I can't find anything about https://github.com/storybookjs/addon-designs support going away

kerryarchibald (approved):
Looks great! Thank you

PR #101, opened 2023-10-03
Enforce the use of logical CSS properties

robintown:
This stylelint plugin even comes with an autofixer, so it was pretty quick to set up.

Closes https://github.com/vector-im/compound/issues/233


germain-gg (approved):
Lovely stuff, thank you 💯

PR #103, opened 2023-10-12
Tooltip: rename shortcut to caption and styling fixes

germain-gg:
Fixes https://github.com/vector-im/compound/issues/246
Following the discussion on https://github.com/vector-im/compound-web/pull/102


t3chguy (changes requested):
https://github.com/vector-im/compound/issues/246 says to use `caption`

t3chguy (changes requested):
Can you update the PR title to include the change from small->span also? Thanks for the compat

t3chguy (approved):
Wrong button

PR #104, opened 2023-10-12
Form: implement read-only state on controls and make the help text better

sandhose:
This implements the `readonly` state on inputs as seen in MAS designs: https://www.figma.com/file/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?type=design&node-id=3464-13893&mode=dev

I also polished a bit the form controls, including:

 - fix the margin between the label and the field
 - made the label use the right font
 - made the help text use the right font, color and margin
 - added a `kind="error"|"help"` prop on `<Form.Message />` component, with a warning icon when `kind == "error"`
 - added a story for a single input with the right controls (and hopefully the right percy snapshots)

Please check out the stories here: https://quenting-readonly-control.compound-web.pages.dev/?path=/story/control--basic

### DIFF
--- a/src/components/ActionControl/ActionControl.module.css
+++ b/src/components/ActionControl/ActionControl.module.css
@@ -15,11 +15,12 @@ limitations under the License.
 */
 
 .actioncontrol {
-  max-inline-size: fit-content;
   position: relative;
+  display: inline-flex;
 }
 
 input.input {
+  flex: 1;
   padding-inline-end: var(--cpd-space-12x);
 }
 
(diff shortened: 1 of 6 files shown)


kerryarchibald (commented):
What is the difference in use case for a disabled and read only control? Should a readonly control still be focusable?

germain-gg:
Looks like you're spot on Kerry, https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly#attribute_interactions

@janogarcia flagging this pull request as we need to spec out the `readonly` state in the Figma Web Components library.

sandhose:
Closed in favour of #108

PR #111, opened 2023-10-18
Replace start with flex-start in Alert styles

weeman1337:
There should be no visual change

fixes https://github.com/vector-im/compound/issues/134

### DIFF
--- a/src/components/Alert/Alert.module.css
+++ b/src/components/Alert/Alert.module.css
@@ -18,8 +18,8 @@ limitations under the License.
 
 .alert {
   display: flex;
-  align-items: start;
-  justify-content: start;
+  align-items: flex-start;
+  justify-content: flex-start;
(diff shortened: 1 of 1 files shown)


germain-gg (changes requested):
I do not think this is the right fix. 

[`align-items` `start` and `end`](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items) are well supported on all browsers that Compound claims to support.
If a downstream project has a different set of browser requirements, it should transpile the incoming external CSS rather than requiring the upstream project to change.

A similar issue has been raised a few times in element web, and it does look like `css-loader` is outdated.

If we decide that this change is what we want to have, we need to add a stylelint rule to prevent `start` and `end` to be used, otherwise this issue is going to come back

Looking at https://github.com/vector-im/compound/issues/134 ,

PR #113, opened 2023-10-23
Create separator component

germain-gg:
Initially, we'll use it just for explicit/standalone `<hr>` elements.

There are no immediate plans to replace generic borders in components with standalone, nested divider elements.

### DIFF
--- a/src/components/Separator/Separator.tsx
+++ b/src/components/Separator/Separator.tsx
@@ -1,0 +1,68 @@
+/*
+Copyright 2023 New Vector Ltd
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import classnames from "classnames";
+import * as RadixSeparator from "@radix-ui/react-separator";
+import React, {
+  ComponentProps,
+  PropsWithoutRef,
+  Ref,
+  forwardRef,
+  useEffect,
+} from "react";
+import styles from "./Separator.module.css";
+
+type SeparatorProps = {
+  /**
+   * The CSS class name.
+   */
+  className?: string;
+  /**
+   * The spacing on either side of the separator
+   */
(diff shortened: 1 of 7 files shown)


janogarcia:
Initially, we'll use it just for explicit/standalone <hr> elements.

Unsure about the benefits of replacing most border decorations (for example, in menus) with a nested instance of a divider component/element, as long as we can apply semantic tokens for those border decorations.

janogarcia:
I'll edit the description to reflect that.

src/components/Separator/Separator.tsx
```diff
@@ -0,0 +1,68 @@
...
+      ...props
+    }: PropsWithoutRef<SeparatorProps>,
+    ref?: Ref<HTMLDivElement>,
+  ) => {
+    const classes = classnames(styles.separator, className);
+
+    useEffect(() => {
+      // Casting to `RefObject` as everything here should use this type.
+      // Functions refs are not something we wish to support and string refs
+      // are already unsupported with `forwardRef`.
```
t3chguy, line 55: Please move this doc to component level so consumers can see this without needing to open the source code to avoid very difficult to diagnose bugs.

germain-gg: I'd rather not overload the JSDoc comment. If that's a concern, i will instead throw an exception

t3chguy: Throwing an exception is a runtime protection, we should be leveraging build time protections really

germain-gg: How is that closer? it will still fail silently... i'd rather have a slightly delayed feedback but something a bit more explicit.

I guess i'll do both 👍

germain-gg: Only ended up adding disclaimer in JSDoc to avoid extra complexity with throwing, this required some ninja moves in TypeScript that would the same complexity as the other solution (hook based)

src/components/Separator/Separator.tsx
```diff
@@ -0,0 +1,72 @@
...
```
robintown, line 59: If you're interested, an alternative solution that makes it possible to support function refs would be to create a new `const ownRef = useRef()`, and then [merge it with the forwarded ref](https://github.com/vector-im/element-call/blob/13137e9f3c282cdf0e7124b555b41dcb8f556bfc/src/room/InCallView.tsx#L166)

robintown (approved):

src/components/Separator/Separator.tsx
germain-gg: Hmmm really cool! Didn't know that existed, thank you!

PR #108, opened 2023-10-17
Forms cleanup

sandhose:
This does almost everything layed out in https://github.com/vector-im/compound/issues/249

It also includes #104 and #106 

Main exception is that I did not touch the Radio, Toggle and Search components.

This is best reviewed commit by commit! Happy to split it up in smaller PRs if that's preferred.

PR #115, opened 2023-10-23
Remove Figma definitions in storyboook

germain-gg:
It does not work in newer Storybook versions


sandhose:
That's a shame, I used it a lot while implementing #108!

PR #118, opened 2023-10-29
Update gen component template to use forwardRef

kerryarchibald:
Fixes https://github.com/vector-im/compound/issues/252

Should be enforced with a rule https://github.com/vector-im/compound/issues/258


germain-gg (approved):
Looks good to me!

As a follow up to this pull request, we should probably create an epic that lists all components that use the `as` pattern and replace it with `asChild`

PR #121, opened 2023-10-31
Alert component tweaks

kerryarchibald:
For https://github.com/vector-im/compound/issues/250

- [x] Changed: Set description text to `font.body.sm.regular`
- [ ] Added: New variant with a stacked layout. Actions on the bottom.
- [ ] Added: New variant with a 2 column layout. Actions on the right column, vertically centered. It should shift to stacked layout when the component width (container query) is below 600px.
- [x] Bug: The corner radius should be 8px.
- [x] Bug: The close/dismiss button is not correctly sized.

Not adding new variants in this change.

[Figma — Alert Component](https://www.figma.com/file/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?type=design&node-id=795-7802&mode=design&t=UVgZk0pPlOaqkhma-0)


germain-gg (approved):
Looks good to me 👍

PR #124, opened 2023-11-02
Add Table, FilterChips components

Half-Shot:
We've been using these in ESS for a while, so it feels like the right time to stabilize them inside compound. I think this is mostly sane, but there are a few outstanding areas:

- The dropdown menu replacement needs a lot of work. We've been using bootstrap internally, but properly implement dropdowns here.
- ??

image


dav-is:
> Should the underlying table use [...]

+1 on using Radix for as much as possible

Half-Shot:
> > Should the underlying table use [...]
> 
> +1 on using Radix for as much as possible

Works for me :+1:

Half-Shot:
Hey there people on the tail end of 2025, closing this PR as in practice we seem to have gotten along for the last couple of years without it and I've not had time to focus on it.

PR #125, opened 2023-11-02
Add `style` package export for dist/style.css

kerryarchibald:
The following syntax will set the style condition:

- CSS @import
- HTML <link rel="stylesheet">

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     },
     "./dist/style.css": {
       "require": "./dist/style.css",
-      "import": "./dist/style.css"
+      "import": "./dist/style.css",
+      "style": "./dist/style.css"
     }
   },
   "types": "./dist/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -9,12 +9,12 @@
     ".": {
       "require": "./dist/compound-web.cjs",
       "import": "./dist/compound-web.js",
-      "types": "./dist/index.d.ts"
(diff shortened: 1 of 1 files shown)


dbkr (approved):
In wasn't aware of this but looks like it's a thing that webpack reads (https://webpack.js.org/guides/package-exports/#conditions)? Looks fine. What happened if the style export wasn't set? It failed rather than falling back to one of the others?

kerryarchibald:
> In wasn't aware of this but looks like it's a thing that webpack reads (https://webpack.js.org/guides/package-exports/#conditions)? Looks fine. What happened if the style export wasn't set? It failed rather than falling back to one of the others?

Yep, when the specific type is not specified it will result in module not found:

> When the exports field is specified, only these module requests are available. Any other requests will lead to a ModuleNotFound Error.

package.json
```diff
@@ -13,7 +13,8 @@
     },
     "./dist/style.css": {
       "require": "./dist/style.css",
-      "import": "./dist/style.css"
+      "import": "./dist/style.css",
+      "style": "./dist/style.css"
```
sandhose, line 17: Wouldn't it make sense then to set it in the `.` import? I think that would make `@import "@vector-im/compound-web"` work?

kerryarchibald: Oh good idea!

PR #126, opened 2023-11-06
Forms: make ligatures opt-in on inputs

kerryarchibald:
For https://github.com/vector-im/element-web/issues/25538

Adds `disableLigatures` prop to `Form.Text`
Default behaviour is for ligatures to match the rest of typography, ie enabled.

See:
- https://57d2e811.compound-web.pages.dev/?path=/story/typography--ligatures
- https://57d2e811.compound-web.pages.dev/?path=/story/form-controls-text--filled

### DIFF
--- a/src/components/Alert/__snapshots__/Alert.test.tsx.snap
+++ b/src/components/Alert/__snapshots__/Alert.test.tsx.snap
@@ -37,12 +37,12 @@ exports[`Alert > renders actions 1`] = `
         class="_text-content_6cfd7b"
       >
         <p
-          class="_font-body-md-semibold_489030"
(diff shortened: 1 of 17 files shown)


weeman1337 (approved):
Looks good !

t3chguy:
Hmm, IMO this should be more specific and less global, any input field which ends up being "sent" and then drawn like room name, topic, composer, etc would be confusing if it rendered differently

kerryarchibald:
@t3chguy fair point, i've changed this to be a per-input `disableLigatures` property

t3chguy:
@kerryarchibald great, what do you think about opt-in vs out?

PR #136, opened 2024-01-04
Avoid conditional export for `./dist/style.css`

richvdh:
For some reason that I don't entirely understand, the conditional export here seems to upset postcss-loader (https://github.com/element-hq/element-web/issues/23595#issuecomment-1877670351) -- possibly because this is neither a cjs nor an es6 import. Since both `require` and `import` point to the same file, we might as well make it unconditional.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,10 +12,7 @@
       "types": "./dist/index.d.ts",
       "style": "./dist/style.css"
     },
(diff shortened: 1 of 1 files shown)


robintown (approved):
Hmm, looks plausible?

PR #140, opened 2024-01-10
Refactor IndicatorIcon out of IconButton

dbkr:
So we have an icon-with-indicator component that we cvan use without it being a button.

This also updates the possible colours for the indicator which we need for the new threads notification work. I've renamed them to match the colours they use rather than 'highlight' as previously because I found it the least confusing. This does mean that this is a BREAKING CHANGE (although we could add 'highlight' for backwards compat, I don't think it's necessary?)


src/components/Alert/__snapshots__/Alert.test.tsx.snap
```diff
@@ -77,19 +77,24 @@ exports[`Alert > renders actions 1`] = `
...
       tabindex="0"
     >
-      <svg
-        class="cpd-icon"
-        fill="none"
-        height="1em"
-        viewBox="0 0 24 24"
-        width="1em"
-        xmlns="http://www.w3.org/2000/svg"
+      <div
```
dbkr, line 80: Almost all of the snapshot updates are adding a div for the extra component.

src/components/Button/IconButton/IconButton.module.css
```diff
@@ -21,50 +21,23 @@ limitations under the License.
 
   inline-size: var(--cpd-icon-button-size);
   block-size: var(--cpd-icon-button-size);
-  padding: calc(var(--cpd-icon-button-size) * 0.125);
+  padding: 0px;
```
dbkr, line 24: I don't really understand why this padding was here and it was throwing things off. Any insights welcome but as far as I can see it was unnecessary.

dbkr: Never mind, see second commit.

src/components/Icon/IndicatorIcon/IndicatorIcon.module.css
```diff
@@ -0,0 +1,68 @@
...
+  mask-size: 100%;
+}
+
+.indicator-icon[data-indicator]::before {
+  content: "";
+  position: absolute;
+
+  /* The dot is 1/12th above & to the right of the canvas */
+  inset-block-start: -8.3333%;
+  inset-inline-end: -8.3333%;
```
dbkr, line 53: The linter complains if we use top / right here, but do we really want the dot on the top left if the user is using an ltr script?

florianduros: To be honest, I don't know at all. The indicators are flipped in the case of a ltr ?

src/components/Alert/__snapshots__/Alert.test.tsx.snap
florianduros: If the `IndicatorIcon` is disabled, it's possible to not render it ? We adding an extra div in the case of the alert which is not useful

dbkr: I tried with a `<>` but I think we need something to apply the styles to in the case where its not a button.

src/components/Icon/IndicatorIcon/IndicatorIcon.module.css
dbkr: I hadn't tested but, yep:

Screenshot 2024-01-23 at 17 27 42

florianduros: huhu, it doesn't look good :(
Do we have also the issue on `main` with the `IconButton`?

src/components/Alert/__snapshots__/Alert.test.tsx.snap
florianduros: Oh okay

src/components/Icon/IndicatorIcon/IndicatorIcon.module.css
dbkr: Same applies before, for the record: this is on current nightly with dir="rtl":

Screenshot 2024-01-23 at 17 36 02

florianduros: Ok, I think it's not a blocker for this PR

florianduros (approved):

PR #141, opened 2024-01-15
`Icon` in `MenuItem` can be a `ReactNode`

florianduros:
This PR allows to use a custom component as an `Icon` in the `MenuItem`.

For example, a nice teal rectangle:
image

### DIFF
--- a/src/components/Menu/MenuItem.tsx
+++ b/src/components/Menu/MenuItem.tsx
@@ -19,6 +19,8 @@ import React, {
   ComponentPropsWithoutRef,
   ComponentType,
   ElementType,
+  isValidElement,
+  ReactNode,
   SVGAttributes,
   useCallback,
   useContext,
@@ -43,7 +45,7 @@ type Props<C extends MenuItemElement> = {
   /**
    * The icon to show on this menu item.
    */
-  Icon: ComponentType<SVGAttributes<SVGElement>>;
+  Icon: ComponentType<SVGAttributes<SVGElement>> | ReactNode;
   /**
    * The label to show on this menu item.
    */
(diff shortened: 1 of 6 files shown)


src/components/Menu/MenuItem.tsx
```diff
@@ -113,7 +119,17 @@ export const MenuItem = <C extends MenuItemElement = "button">({
       onClick={onClick}
       disabled={disabled}
     >
-      <Icon width={24} height={24} className={styles.icon} aria-hidden={true} />
+      {iconIsReactElement ? (
+        <div className={styles.icon}>{componentIcon}</div>
```
t3chguy, line 123: Why not use Radix SlotClone instead of bloating the DOM?

florianduros: Because I never head of Radix SlotClone :)

t3chguy: Yes, its the basis for how all of the Radix stuff Compound builds upon works

florianduros: This utility is not currently installed in Compound. Are you talking about the `asChild` pattern ?

florianduros: I tried with `Slot`:
- It works if the children is directly JSX like: `<Slot classname="myClass"><span>content</span></Slot>`
- It doesn't work if I'm doing `<Slot classname="myClass">{myComponent}</Slot>`

t3chguy: It empirically works in Tooltip

florianduros: I made it works, I forgot to spread the props of the test component 🙃

t3chguy (approved):

PR #145, opened 2024-01-27
Switch to Playwright for visual testing

robintown:
Following the lead of Element Web, I propose we switch to using Playwright for visual testing, rather than Percy. (Context: Percy has been non-functional on this repository for months since we kept going over our billing limits.) This means that screenshots will now be stored directly in this repository as Playwright snapshots, which should be a much more cost-effective solution.

Note that this change also includes a couple of fixes for broken stories that the tests caught (yay!) The "Kitchen Sink" form stories were crashing due to lack of a TooltipProvider, and the search component was rendering inconsistently on dev vs. production builds due to bits of CSS being emitted in a different order.

Closes https://github.com/element-hq/compound/issues/278


t3chguy (approved):
Looks generally sane, you should probably include a Docker command for updating the snapshots for those not on Linux, otherwise you're going to make it difficult for contributors

PR #147, opened 2024-02-06
Fix `Menu` focus after closing

florianduros:
In https://github.com/matrix-org/matrix-react-sdk/pull/12136 , we are building the TAC (Threads Activity Centre) with the `Menu` component. We have the current issue: when the TAC is closing, the button is the Spaces has his tooltip displayed again. (You can see the tooltip `Fils de discussion` at the bottom left)

https://github.com/element-hq/compound-web/assets/2621378/7cac0750-5581-4d2c-ae23-3642c2eaecbc

Because the focus is set on the `Menu` trigger when the `Menu` is closing. According to the https://www.radix-ui.com/primitives/docs/components/dropdown-menu#content `onCloseAutoFocus` property:

> Event handler called when focus moves to the trigger after closing. It can be prevented by calling `event.preventDefault`.

Adding `event.preventDefault()` stops the `Menu` trigger to have the focus after the `Menu` closing

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -119,7 +119,17 @@ export const Menu: FC<Props> = ({
     <Root open={open} onOpenChange={onOpenChange}>
       <Trigger asChild>{trigger}</Trigger>
       <Portal>
-        <Content asChild side={side} align={align} sideOffset={8}>
+        <Content
+          asChild
+          side={side}
+          align={align}
+          sideOffset={8}
+          onCloseAutoFocus={(event) => {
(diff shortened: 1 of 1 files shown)


dbkr (approved):

t3chguy:
This change seems to go against the WAI ARIA guidelines which say

> Close the menu that contains focus and return focus to the element or context, e.g., menu button or parent menuitem, from which the menu was opened.

https://www.w3.org/WAI/ARIA/apg/patterns/menubar/

robintown:
Agreed, I think the correct solution would instead be to apply a fix like this: https://github.com/matrix-org/matrix-react-sdk/pull/8546

PR #149, opened 2024-02-07
Change how we align the chevron on a MenuItem

dbkr:
Hopefully the comment explains. The chevron SVG has a lot of blank space and so needs shoving over to get it close enough to the right hand edge. We did this with a negative margin but this causes some of the chevron element to lie outside the parent which can make scrollbars appear and/or cause it to cover the focus-active border.

This adjusts the viewBox of the SVG so it can just be the right size for the space.

Fixes a bug where horizontal scrollbars would appear on items in the TAC as they were hovered over in chrome.

Screenshots changed because there's now more space for text. This seems desirable I think?


playwright/visual.test.ts-snapshots/Menu-MenuItem-With-A-Long-Label-And-Children-1-chromium-linux.png
robintown, line 1: Huh, this doesn't look right even before the change (nothing to worry about for this PR)

robintown (approved):

PR #152, opened 2024-02-26
Fix the action input overflowing of small containers

sandhose:
This is one of the weirdness of flexbox. See https://dfmcphee.com/flex-items-and-min-width-0/

Before: ([Storybook](https://compound-web.pages.dev/?path=/story/form-kitchen-sink--normal)) 
image

After: ([Storybook](https://quenting-action-input-min-wi.compound-web.pages.dev/?path=/story/form-kitchen-sink--normal)) 
image


t3chguy (approved):
Never seen `min-inline-size` - EW normally uses `min-width` for the same trick but its probs fine

PR #156, opened 2024-04-09
Tooltip rework

florianduros:
Closes https://github.com/element-hq/compound/issues/298

Migrate the current tooltip implementation from radix to floating-ui.

The tooltip props has changed, the next release should be a major release:
- `side`, `align` & `placement`
    - radix-ui is using two different props for positioning: `side` and `align`
    - floating-ui is using one prop: `placement` which is basically `side` and `align` merged in one props
- `onEscapeKeyDown` & `onPointerDownOutside` are partially replaced by `onOpenChange`. Currently they are not used in the `react-sdk`.


playwright/visual.test.ts-snapshots/Form-Controls-Password-Visible-1-chromium-linux.png
florianduros, line 1: The focus (blue circle) is on the eye button, so the tooltip should appear, right ?

dbkr: Seems reasonable, yeah

dbkr (approved):
Phew, thanks for doing this - certainly looks like an improvement at least: nice that it doesn't `<TooltipProvider>`s everywhwere. Lets hope it works out better too.

PR #160, opened 2024-04-19
Fix Storybook link

robintown:


t3chguy (approved):

florianduros:
The url is also here https://github.com/element-hq/compound-web/blob/main/.storybook/ElementTheme.ts#L18

PR #166, opened 2024-05-14
Edit In Place component

dbkr:
A component that allows a user to edit text in place with buttons that reveal while editing.


src/components/Form/Controls/EditInPlace/EditInPlace.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+    const onSaveButonClicked = useCallback(async () => {
+      if (saveDisabled) return;
+
+      try {
+        onSave();
+        setShowSaved(true);
+        setTimeout(() => {
+          setShowSaved(false);
+        }, 2000);
```
robintown, line 125: I believe this will log a warning if the component is unmounted while the timer is still running

robintown (commented):
Honestly, I don't see anything wrong with keeping `showSaved` internal to the component. Controlled components are preferable to uncontrolled components when the parent might reasonably want to have access the child's state, but I don't think that applies to the "saved" text here - it just isn't useful to know about.

florianduros (commented):
It will be great to have a playwright snapshot with the save and cancel buttons

florianduros (approved):

PR #167, opened 2024-05-14
Memoize cloned element in tooltip

florianduros:


src/components/Tooltip/Tooltip.tsx
```diff
@@ -163,15 +164,21 @@ function TooltipAnchor({ children }: Readonly<PropsWithChildren>): JSX.Element {
...
+    if (!isValidElement(children)) return;
+
+    const props =
+      isValidElement(children) &&
+      context.getReferenceProps({
+        ref,
+        ...children.props,
+      });
+    return cloneElement(children, props);
+  }, [ref, children, isValidElement(children) && children.props]);
```
t3chguy, line 177: `&&` here is unexpected, it implies breaking rules-of-hooks

florianduros: Why is it unexpected?

t3chguy: Looks like this project doesn't use https://github.com/facebook/react/tree/main/packages/eslint-plugin-react-hooks which has rules-of-hooks. https://www.reddit.com/r/reactjs/comments/13oy1r5/is_eslint_exhaustive_deps_a_bad_rule_sometimes/

florianduros: Oh I see. I added a comment. It's needed to be able to infer the type of children to have access to the `props` field.

t3chguy: I was more questioning why `&&` instead of `,` - a comment or maybe breaking it out into a sanely named variable could go a long way to clarifying it

t3chguy (approved):

PR #168, opened 2024-05-14
Fetch avatars without CORS

robintown:
By omitting the `crossOrigin` attribute, we change avatars to use a non-CORS request to fetch the image. This allows avatars to load even on null origins (fixing https://github.com/element-hq/element-web/issues/26491), but also prevents us from inspecting the image data in JavaScript, for example by using it in a canvas. We weren't doing that of course, so switching to the locked-down non-CORS mode should have no consequences.

Closes https://github.com/element-hq/compound/issues/293


t3chguy (approved):

t3chguy:
thanks

PR #170, opened 2024-05-15
Revert "Memoize cloned element in tooltip"

florianduros:
Reverts element-hq/compound-web#167

It's create regression in rerender of the anchor.

We should do a deep comparison of the new computed props and use it in the memo dependencies.


t3chguy (approved):

florianduros:
Closing in favour of https://github.com/element-hq/compound-web/pull/171

PR #172, opened 2024-05-15
Major dependency updates

dbkr:
* Storybook 8
 * Vite 5
 * Vitest 1

There are *a lot* of changes from Storybook 7 to 8. I believe this should work. The migration tool helped (sorta) but if you spot anything broken, it could be this.

Removes storybook-addon-jsx because it doesn't appear tgo be compatible with storybook 8 (uses @storybook/addons) and hasn't been updated since 2021.


.node-version
```diff
@@ -0,0 +1 @@
+20.13.1
```
dbkr, line 1: This bumps the node version that cloudflare pages uses. It's a shame we can't comment this file. :/

robintown (approved):
Thanks a bunch for doing this!

PR #173, opened 2024-05-24
Implement the Toast component

dbkr:


src/components/Toast/Toast.module.css
```diff
@@ -0,0 +1,25 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+.toast-container {
+  inline-size: fit-content;
+  background-color: var(--cpd-color-alpha-gray-1300);
+  color: var(--cpd-color-text-on-solid-primary);
+  border-radius: 99px;
+  font-size: 13px;
```
florianduros, line 22: Can we use the compound token instead?

florianduros (approved):

src/components/Toast/Toast.module.css
dbkr: ah, good point.

PR #174, opened 2024-05-28
Make title optional on Menu

dbkr:


t3chguy (commented):
I feel like if a title is not provided then a label must be, can the types reflect this?

dbkr:
Not sure what you mean... there is no prop for label, or were you suggesting adding one?

src/components/Menu/FloatingMenu.tsx
```diff
@@ -41,25 +56,18 @@ interface Props extends ComponentPropsWithoutRef<"div"> {
...
-  ({ title, className, children, ...props }, ref) => {
+  ({ title, showTitle = true, className, children, ...props }, ref) => {
     const titleId = useId();
     return (
       <div
         role="menu"
         ref={ref}
-        aria-labelledby={titleId}
+        aria-label={showTitle ? undefined : title}
+        aria-labelledby={title ? titleId : undefined}
```
t3chguy, line 66:
```suggestion
        aria-labelledby={showTitle ? titleId : undefined}
```

t3chguy (changes requested):

PR #175, opened 2024-05-30
Make EditInPlace submit on enter

dbkr:
...by wrapping it in a form


t3chguy (commented):
I'm not sure I agree with the approach, wouldn't this be a component within a larger form for correct a11y semantics?

dbkr:
As I understand it, it wouldn't, since the point of the the edit in place control is that the save button commits that change individually. I think it would be incorrect to use it as part of a larger form since it would still submit separately.

t3chguy:
Then I think we need docs to state that it must not be used within a form given it has its own form

t3chguy (approved):

PR #177, opened 2024-06-05
Improve Search onChange signature

langleyd:
use `React.ChangeEvent<HTMLInputElement>` in the onChange signature of the `Search` component to avoid having to cast it in the application layer.


src/components/Search/Search.test.tsx
```diff
@@ -30,4 +41,22 @@ describe("Search", () => {
...
+    const onChange = vi
+      .fn()
+      .mockImplementation((e: ChangeEvent<HTMLInputElement>) => {
+        value = e.target.value;
+      });
+    render(<SearchTest onChange={onChange} />);
+    const query = "my query";
+    await act(async () => {
+      const input = screen.getByRole("searchbox");
+      await userEvent.type(input, query);
```
dbkr, line 56: As far as I understand, the user-event library handles this so it doesn't need to be wrapped in an act(). At least, user-event's examples don't use act(): https://testing-library.com/docs/user-event/intro/#writing-tests-with-userevent

dbkr (approved):

PR #181, opened 2024-06-13
replace graphemer by unicode-segmenter

cometkim:
I made a Unicode library that is much smaller and faster than graphemer. Check it out: https://github.com/cometkim/unicode-segmenter?tab=readme-ov-file#unicode-segmentergrapheme-vs

- 2x smaller
- 6~9x faster
- ESM/CJS support
- Latest Unicode data

It ensures compliance with the latest Unicode data by performing tests and fuzzing against the `Intl.Segmenter` API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "@radix-ui/react-tooltip": "^1.0.6",
     "@storybook/manager-api": "^8.1.1",
     "classnames": "^2.3.2",
-    "graphemer": "^1.4.0",
+    "unicode-segmenter": "^0.9.0",
     "vaul": "^0.7.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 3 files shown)


dbkr:
Looks like author has mass-created the same pull request to every dependant of graphemer. I'm treating this as spam, and likely to be malicious.

t3chguy:
I think it isn't malice but over-excitement. But in the case of this repo I think the usage is so minimal that maybe going straight for Intl.Segmenter is the way to go, will keep this on my list to look into

cometkim:
> Looks like author has mass-created the same pull request to every dependant of graphemer. I'm treating this as spam, and likely to be malicious.

It hurts, I understand your doubts about this. But you can close this or a legitimate challenge instead of driving me as a spammer...

cometkim:
As I already mentioned, if you can use `Intl.Segment`, it would be better for the long term.

Unfortunately I couldn't do that in my production. And I thought a lot of people would want the same alternative, but I was not good at how to make it known

PR #183, opened 2024-06-18
Add help label to edit in place

florianduros:
Add a help message according to the figma design https://www.figma.com/design/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?node-id=4335-2016&t=m2StPHnu3yvpsXaj-4

image


src/components/Form/Controls/EditInPlace/EditInPlace.tsx
```diff
@@ -125,6 +131,10 @@ export const EditInPlace = forwardRef<HTMLInputElement, Props>(
       [styles["container-show-buttons"]]: saving,
     });
 
+    // Display the help label if there is no other labels
+    const displayHelpLabel =
+      Boolean(helpLabel) && !error && !saving && !(savedLabel && showSaved);
```
robintown, line 136: This feels brittle, could we instead try something like:

```typescript
const labels: JSX.Element[] = []
if (error) labels.push(...)
if (saving) labels.push(...)
if (labels.length === 0 && helpLabel) labels.push(...)

File: src/components/Form/Controls/EditInPlace/EditInPlace.tsx [hunk 2]
```diff
@@ -215,70 +211,138 @@ export const EditInPlace = forwardRef<HTMLInputElement, Props>(
...
+function Labels({
+  error,
+  errorTextId,
+  saving,
+  savingLabel,
+  savedLabel,
+  showSaved,
+  helpLabel,
+}: LabelsProps): JSX.Element {
+  const labels: JSX.Element[] = [];
```
robintown, line 271: Each element in the list will need a `key` prop (unfortunately React can only warn about this at runtime)

src/components/Form/Controls/EditInPlace/EditInPlace.tsx
```diff
@@ -205,59 +211,139 @@ export const EditInPlace = forwardRef<HTMLInputElement, Props>(
...
+  if (error) {
+    labels.push(
+      <div className={styles["caption-line"]} key={error}>
```
robintown, line 274: This is probably okay, but I would go for static keys like `key="error"` because technically nothing is stopping two labels from having the same text. It also allows React to produce a more fine-grained diff when the text of a label changes.

robintown (approved):

PR #188, opened 2024-06-19
Refactor EditInPlace

sandhose:
This refactors the EditInPlace component to reuse a lot of the existing styling in existing controls

It solves a bunch of problem:

 - there was no tooltip
 - styling was mostly duplicated and inconsistent, especially on the buttons
 - the component would try to grow instead of keeping its natural size
 - race condition in the 'saving' timer
 - this uses regular for onSubmit/onReset instead of relying on onClicks on the buttons
 - all the aria labels are managed by radix forms
 - the state is actually managed by a somewhat sensible state machine instead of a bunch of boolean states, less likely to go in unknown states
 - the component would only work if you used a 'controlled' input
 - the tests had a nested form for some reason?
 - the visual screenshots actually did not include the edit buttons
 - onSave did not pass on the form event, so no way to get the input value that way

A few breaking changes on the API:

 - all the labels are now mandatory
 - `onSave`/`onCancel` pass on the form event, use that to grab the formData
 - `disableSaveButton` has been removed, but probably not needed as the component now detects when things have changed or not


dbkr:
Okay, this is quite sizeable. I'd just finished getting this component right for element-web and it was quite subtle, although admittedly not the cleanest. I will try to have a proper look at this in the near future, but some initial points:

 * Will the field still be the full width of the form in the dialog, which it needs to be?
 * Web uses the disableSaveButton to disable saving if the display name is the same as the original, including being edited and then changed back - I would guess this is also fine, but it's a change.
 * You've tested it in element-web with the interface changes?

t3chguy (approved):
Looks sane

sandhose:
> Will the field still be the full width of the form in the dialog, which it needs to be?

The form will now adopt its natural size, like all the other controls, so if it's wrapped in a flexbox/grid, it will take the right size.

> Web uses the disableSaveButton to disable saving if the display name is the same as the original, including being edited and then changed back - I would guess this is also fine, but it's a change.

If you *can* avoid using a controlled input, e.g. use `defaultValue` and rely on the form data you get in `onSave`, I think you would get a more robust implementation.
If not, I can introduce a `dirty` flag, which if set would let the parent component decide whether the form is considered dirty or not.

> You've tested it in element-web with the interface changes?

I haven't, I don't have a local dev setup for element-web, and am mainly interested for MAS, as we already have a home-grown, simpler version of this component

sandhose:
@dbkr do you want to still have a look or are you fine with @t3chguy green tick?

dbkr:
Go for it - from my first look it looked sane.

PR #190, opened 2024-06-19
InlineSpinner: fix forward ref

florianduros:
Closes https://github.com/element-hq/compound-web/issues/187

### DIFF
--- a/src/components/InlineSpinner/InlineSpinner.tsx
+++ b/src/components/InlineSpinner/InlineSpinner.tsx
@@ -14,21 +14,20 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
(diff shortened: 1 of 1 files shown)


t3chguy (commented):
What if the caller does need a ref on the spinner?

florianduros:
> What if the caller does need a ref on the spinner?

When I try to use a forwared ref on it, I have the following error: `Property ref does not exist on type: IntrinsicAttributes & SVGAttributes<SVGElement>`

t3chguy:
Sounds like the svg loader needs fixing to support refs

dbkr (approved):

florianduros:
@t3chguy created an issue to track it https://github.com/element-hq/compound-web/issues/192

PR #193, opened 2024-06-20
Inline the avatar mark to remove the react-dom/server dependency

sandhose:
This unnecessarily adds a react renderer to the bundle/dependencies just for a simple SVG that doesn't change.

It's also a step towards removing SVGR from this project.


t3chguy (commented):
This is a step towards making it harder to simply diff files between Figma and git, or better yet automate them for design to own them

sandhose:
> This is a step towards making it harder to simply diff files between Figma and git, or better yet automate them for design to own them

I would agree for icons in the design tokens, but it feels like this path was manually crafted?

t3chguy:
React-SDK had a similar mask (before switching to Compound Avatars) which came from Figma

sandhose:
Fine, I replaced that with a raw import of the SVG. I guess the bundle size would be slightly bigger as the SVG has comments in it, but it's nothing compared to react-dom/server

t3chguy (approved):
Looks sane

PR #194, opened 2024-06-20
Use the icon components directly from the tokens package

sandhose:
This uses the icons directly exported by `compound-design-tokens`, which means we can remove the `svgr` plugin/logic here.

It also includes a few quality of life improvement, like the fact that stories which have icons in them now have *all* the icons instead of just a few


vite.config.ts
```diff
@@ -24,21 +23,20 @@ export default defineConfig({
...
         "vaul",
         "@floating-ui/react",
-        "@floating-ui/react-dom",
         "@radix-ui/react-context-menu",
         "@radix-ui/react-dropdown-menu",
         "@radix-ui/react-form",
         "@radix-ui/react-separator",
         "@radix-ui/react-slot",
         "@radix-ui/react-tooltip",
+        /^@vector-im\/compound-design-tokens\/.*/,
```
robintown, line 39: Can you comment on why this isn't just `"@vector-im/compound-design-tokens"`?

src/components/Icon/Gallery.mdx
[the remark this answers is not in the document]
sandhose: Oh it is! Reverted

vite.config.ts
sandhose: I added a comment, hopefully clear enough?

robintown (approved):
Right, looks good!

PR #195, opened 2024-06-21
Add label and destructive props to `IconButton`

florianduros:
Currently, we need to manually add the destructive style every time we want to reproduce the destructive design on icon.
Also, add an optional tooltip to a display label. I think in most of the case when we are displaying an IconButton we will want to display tooltip.

Figma https://www.figma.com/design/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?node-id=656-7159&t=LQkeRLdK2REfOjOX-0


src/components/Button/IconButton/IconButton.tsx
```diff
@@ -39,11 +40,19 @@ type IconButtonProps = UnstyledButtonPropsFor<"button"> &
...
     disabled?: boolean;
+    /**
+     * Whether this button triggers a destructive action.
+     * @default false
+     */
+    destructive?: boolean;
+    /**
+     * The label for the button
+     */
+    label?: string;
```
t3chguy, line 55: the prop says its a label, in reality its a tooltip? Why can't the caller just wrap it in the tooltip like most other components do

florianduros: Because I think that the pair tooltip and `IconButton` is very common. I add it to avoid to add it manually every time. I want to make it easier/less painful to use.

florianduros: @t3chguy You think that this tooltip is too much?

t3chguy: No I just think the docs should say its a tooltip, otherwise the caller has no way to know their label won't just be a label, but be a tooltip

florianduros: Yep make sense, I updated it

src/components/Button/IconButton/IconButton.tsx
```diff
@@ -50,9 +50,9 @@ type IconButtonProps = UnstyledButtonPropsFor<"button"> &
      */
     destructive?: boolean;
     /**
-     * The label for the button
+     * Display a tooltip when the button is hovered
```
t3chguy, line 53: I think `is hovered` is misleading here, because the tooltip will also show for keyboard accessibility users which have no pointer/hover means

florianduros: Rhaaaa, what do you propose instead?

t3chguy: Don't need to specify the semantics of when it'll show;

`Optional tooltip for the button`

t3chguy (approved):

PR #198, opened 2024-06-27
EditInPlace: pass error messages as children to allow better messages

sandhose:
Radix forms is supposed to leverage native validation APIs.
This means that setting `required`, `type=email`, `minlength` on the input field should just work.
It already worked, but there was no way to set an error message.

This is because you're supposed to have `<Message match="...">` inside the field, to render the error message for each kind of error you'd expect.

This makes it so that `children` passed in the `<EditInPlace>` component are rendered within the `<Field>`, which lets us do something like:

```jsx
<EditInPlace type="email" required minLength={5} maxLength={100}>
    <ErrorMessage match="valueMissing">This field is required</ErrorMessage>
    <ErrorMessage match="typeMismatch">This field must be an email</ErrorMessage>
    <ErrorMessage match="tooShort">This field must be at least 5 characters long</ErrorMessage>
    <ErrorMessage match="tooLong">This field must be at most 100 characters long</ErrorMessage>
    {/* and for custom errors: */}
    <ErrorMessage match={e => e.endsWith("@matrix.org")}>The email must end with @matrix.org</ErrorMessage>
</EditInPlace>
```

This means we ditch the `error` prop, in favor of the `children` and the `serverInvalid` prop, if you need to report errors from the server.

```jsx
<EditInPlace serverInvalid={Boolean(error)}>
    {error && <ErrorMessage>{error}</ErrorMessage>}
</EditInPlace>
```

I can keep the (deprecated) `error` prop if we want to avoid making a breaking change

### DIFF
--- a/src/components/Form/Controls/EditInPlace/EditInPlace.stories.tsx
+++ b/src/components/Form/Controls/EditInPlace/EditInPlace.stories.tsx
@@ -19,6 +19,7 @@ import React from "react";
 import { EditInPlace } from "./";
 import { Meta, StoryObj } from "@storybook/react";
 import { expect, userEvent, within } from "@storybook/test";
+import { ErrorMessage } from "../../Message";
 
 type Props = { invalid?: boolean } & React.ComponentProps<typeof EditInPlace>;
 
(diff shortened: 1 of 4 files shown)


sandhose:
(Sorry for the few pushes, I got scared by Sonar telling me test coverage was going done, but now it's actually ready for review)

dbkr (approved):
oh, yay for native form validation

PR #201, opened 2024-07-03
Move ts-xor to dependencies

MidhunSureshR:


dbkr:
is this because we always used it as a dependency or because you'd like to?

MidhunSureshR:
This is to fix https://github.com/element-hq/compound-web/pull/94 which incorrectly mentioned ts-xor as dev dependency leading to https://github.com/matrix-org/matrix-react-sdk/actions/runs/9773852203/job/26980852350?pr=12672#step:5:9

dbkr (approved):

PR #203, opened 2024-07-04
Be tolerant of missing navigator.userAgent / wait for it to become available in SSR environments.

Half-Shot:
Hopefully fixes https://github.com/element-hq/compound/issues/314

### DIFF
--- a/src/components/Menu/ContextMenu.tsx
+++ b/src/components/Menu/ContextMenu.tsx
@@ -111,11 +111,6 @@ export const ContextMenu: FC<Props> = ({
     </Trigger>
   );
 
-  // Platform hasn't loaded yet.
-  if (!platform) {
-    return null;
-  }
-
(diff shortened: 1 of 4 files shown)


src/components/Menu/ContextMenu.tsx
```diff
@@ -88,7 +88,7 @@ export const ContextMenu: FC<Props> = ({
 
   // Normally, the menu takes the form of a floating box. But on Android and
   // iOS, the menu should morph into a drawer
-  const platform = getPlatform();
+  const platform = useMemo(() => getPlatform(), []);
```
t3chguy, line 91: given this pattern is used in multiple places, it seems like a `usePlatform` hook would be nice

t3chguy:
You likely want to reference `navigator` indirectly also

t3chguy (approved):
Looks sane to me

PR #205, opened 2024-07-09
Fix incorrect styling of tab item

MidhunSureshR:
- When used as tabs, we add the `aria-selected` attribute instead of `aria-current`
- Change font weight to 500


MidhunSureshR:
Sorry broken snapshots!

src/components/Nav/Nav.module.css
```diff
@@ -66,6 +66,7 @@
   background: transparent;
   border: 0;
   font: var(--cpd-font-body-md-regular);
+  font-weight: 500;
```
t3chguy, line 69: is there no var for this? Surely CPD should have the right weight set for `Web/font/body/md/medium`

https://github.com/element-hq/compound-design-tokens/blob/240fb7dba469eccc963472033b489ea9be9b1fbf/assets/web/css/cpd-common-base.css#L53
https://github.com/element-hq/compound-design-tokens/blob/240fb7dba469eccc963472033b489ea9be9b1fbf/assets/web/css/cpd-common-base.css#L23

src/components/Nav/Nav.module.css
```diff
@@ -66,6 +66,7 @@
   background: transparent;
   border: 0;
   font: var(--cpd-font-body-md-regular);
```
t3chguy, line 68: Ah, yes, this is wrong should be `medium` not `regular`.

image

t3chguy (approved):

PR #206, opened 2024-07-10
Implement the progress bar component

sandhose:
This also includes a story which shows how to use it for a password strength indicator.

Design reference: https://www.figma.com/design/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?node-id=5425-6829&t=vGtJ0aBTjajPgRzf-4


package.json
```diff
@@ -129,5 +130,6 @@
   "resolutions": {
     "jackspeak": "2.1.1",
     "@radix-ui/react-id": "1.1.0"
-  }
+  },
+  "packageManager": "yarn@1.22.19+sha1.4ba7fc5c6e704fce2066ecbfb0b0d8976fe62447"
```
t3chguy, line 134: This seems a little too precise, what if I have 

```
 ~  yarn --version                                                                                                                                 479ms  Wed 10 Jul 13:33:52 2024
1.22.22
```

Also seems unrelated to this PR

src/components/Progress/Progress.module.css
```diff
@@ -0,0 +1,87 @@
...
+    var(--cpd-progress-bar-main) 50%,
+    var(--cpd-progress-bar-muted) 50%,
+    var(--cpd-progress-bar-muted) 75%,
+    var(--cpd-progress-bar-main) 75%,
+    var(--cpd-progress-bar-main) 100%,
+    var(--cpd-progress-bar-muted) 100%
+  );
+
+  /* sqrt(number of stripes * 2 * (stripe width)^2) = sqrt(4 * 2 * 2^2) = sqrt(32) */
+  background-size: 5.6569px 5.6569px;
```
t3chguy, line 86: That is hellish

src/components/Progress/Progress.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+import { Root, Indicator } from "@radix-ui/react-progress";
+import useId from "../../utils/useId";
+
+type ProgressProps = {
+  /* The size variant of the progress bar */
+  size: "sm" | "lg";
+
+  /* The color variant to use for the progres bar indicator */
+  color?: "red" | "orange" | "lime" | "green";
```
t3chguy, line 30: What happens if you don't specify a colour?

Also we're a British company, colour*
image

and `progress*`

package.json
sandhose: grr, that's my local yarn which added this itself for some reason :(
I think it's corepack related or something?

src/components/Progress/Progress.module.css
sandhose: I'm very sorry about this :(

src/components/Progress/Progress.tsx
sandhose: It becomes gray. You probably should only use this for empty state, as this also propagates to the label on top. See [the password strength story](https://quenting-progress-bar.compound-web.pages.dev/?path=/docs/form-password-form--docs#empty)

And damn I never noticed the britishness of props. But why are the token using `color` then? 🤔 

Also, https://github.com/hashanp/postcss-spiffing

t3chguy: > It becomes gray. You probably should only use this for empty state, as this also propagates to the label on top.

Can the default be commented plz

> But why are the token using color then?

No idea

> https://github.com/hashanp/postcss-spiffing

This is awesome lol

sandhose: I'm avoiding the problem by renaming the prop `tint` if you're alright with that 😬 

I also tweaked the comments, and made sure they get parsed as JSDoc comments, so that they render in the IDE and Storybook

package.json
sandhose: Removed

t3chguy (approved):

PR #208, opened 2024-07-12
Allow setting a ref on InlineSpinner

robintown:
We can do this now that we're consuming the custom SVG components exported by the design tokens package.

Closes https://github.com/element-hq/compound-web/issues/192


sandhose:
Actually in the design tokens we're not forwarding the ref to the SVG. I did that in https://github.com/element-hq/compound-design-tokens/pull/92

t3chguy (approved):

PR #211, opened 2024-07-17
Enable tooltips to be used as accessible labels

robintown:
Currently, the only use case that our tooltip component supports is to describe another element. But it's arguably far more common in our products to use tooltips as an accessible label for an icon. I propose we change the default behavior of `<Tooltip label={...}>` to be that of a persistent accessible label, and use `<Tooltip description={...}>` for descriptive tooltips.

I took the liberty of changing the tooltip tests to use targeted assertions rather than snapshots, because in theory this makes for a less noisy test suite that is also better at communicating what properties we actually expect tooltips to have.

Closes https://github.com/element-hq/compound/issues/322


t3chguy (approved):
Looks sane

PR #213, opened 2024-07-17
Fix test warnings

robintown:
Our test output has become pretty noisy and full of warnings about misuse of `act`. This fixes those warnings; see the commits.

### DIFF
--- a/src/components/Form/Controls/EditInPlace/EditInPlace.test.tsx
+++ b/src/components/Form/Controls/EditInPlace/EditInPlace.test.tsx
@@ -121,6 +121,8 @@ describe("EditInPlace", () => {
     expect(asFragment()).toMatchSnapshot();
 
     const input = getByRole("textbox");
+    // We wrap this in act because the expectation fires an 'invalid' event
+    // internally, causing the component to update. (Is this a bug in RTL?)
     act(() => expect(input).toBeInvalid());
 
     // The message should be there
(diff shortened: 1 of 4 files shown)


src/components/Form/Controls/EditInPlace/EditInPlace.test.tsx
```diff
@@ -112,41 +121,30 @@ describe("EditInPlace", () => {
...
-    expect(input).toBeInvalid();
-
-    // The message hasn't showed up yet
-    expect(queryByText("you did not fill this")).not.toBeInTheDocument();
-
-    await act(async () => {
-      // Focus and submit the form
-      await userEvent.keyboard("{tab}{enter}");
-    });
+    act(() => expect(input).toBeInvalid());
```
t3chguy, line 124: This looks like the wrong way to use `act` - it should wrap mutations, not expectations

https://react.dev/reference/react/act#await-act-async-actfn

robintown: Yet this expectation does fire an `invalid` event on the form internally, which causes the component to react and re-render. I don't yet understand whether this is considered a bug and if so, which Testing Library repo I should file it on, but I should at least add a comment here.

t3chguy: that is awful everything =(

t3chguy (approved):

PR #235, opened 2024-08-12
NavItem: add scrolling interactions

t3chguy:


t3chguy:
Element Web designs no longer use these tabs so closing this PR until something pushes it through

PR #254, opened 2024-10-01
Avoid chaining aria-labelledby on tooltips

robintown:
`aria-labelledby` [cannot be chained](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby#benefits_and_drawbacks). So, it's not good practice to be setting `aria-labelledby` on both the anchor and the tooltip element.

The basic issue that I believe initially motivated the chaining was to distinguish the caption from the main tooltip content (by setting both `aria-labelledby` and `aria-describedby` on the tooltip element), so that the caption would not be considered part of the anchor's accessible label. But this was never actually working. The only good solution seems to be setting `aria-labelledby` and `aria-describedby` on the anchor to point directly to the label and caption elements.


robintown:
https://github.com/element-hq/compound-web/pull/255

PR #259, opened 2024-10-07
Make sure the output can be treeshaked

sandhose:
Fixes element-hq/compound#318

This tells rollup/vite to preserve the file structure of the source directory instead of bundling everything into a single file.

This means that as a downstream package, you'll only import what you need, and the rest will be tree-shaken for you.
Compound itself is rather small, but it has a few dependencies that only a handful of components use.
This means that if you only import a few components, it will be much smaller as it won't include `@radix-ui/*` and `@floating-ui/*` stuff.

Here is the diff on a quick hello-world project bundle analysis, which just imports a `Button`:

#### Before
image

Total gzipped size: 89.15kB

#### After
image

Total gzipped size: 46.60kB

---

In MAS, it means it can code-split way better, putting a lot of compound components out of the main bundle:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "6.3.1",
   "description": "Compound components for the Web",
   "type": "module",
-  "main": "./dist/compound-web.js",
-  "module": "./dist/compound-web.js",
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.js",
   "sideEffects": false,
   "exports": {
     ".": {
--- a/package.json
+++ b/package.json
@@ -5,10 +5,11 @@
   "type": "module",
   "main": "./dist/compound-web.js",
   "module": "./dist/compound-web.js",
+  "sideEffects": false,
   "exports": {
     ".": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -5,10 +5,11 @@
   "type": "module",
   "main": "./dist/compound-web.js",
   "module": "./dist/compound-web.js",
+  "sideEffects": false,
```
sandhose, line 8: This tells downstream bundlers that there is no side-effect importing compound, which lets them actually treeshake stuff out

package.json
```diff
@@ -5,10 +5,11 @@
...
   "exports": {
     ".": {
-      "require": "./dist/compound-web.cjs",
-      "import": "./dist/compound-web.js",
+      "require": "./dist/index.cjs",
+      "import": "./dist/index.js",
```
sandhose, line 12: Because we changed the `fileName` pattern, the output files match the input files structure in `src/`

dbkr (approved):
Nice, thanks!

PR #274, opened 2024-11-11
Ensure dropdown is placed above other elements.

Half-Shot:
I *believe* this would ensure dropdowns work correctly when other `absolute` Elements are placed beneath them.

### DIFF
--- a/src/components/Dropdown/Dropdown.module.css
+++ b/src/components/Dropdown/Dropdown.module.css
@@ -88,6 +88,7 @@ limitations under the License.
       padding: 0;
       padding-block-end: var(--cpd-space-4x);
       cursor: pointer;
+      z-index: var(--cpd-dropdown-z-index);
 
       li {
         list-style: none;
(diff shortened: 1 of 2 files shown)


robintown (changes requested):
I've been really happy that our component library has so far avoided the use of `z-index: 999;` hacks, and am reluctant to start using such a rule here. We've managed to make tooltips and menus display correctly with `z-index: auto;` by putting their "floating" bits in a React portal which gets the right layering by virtue of its location in the DOM. I would prefer to try the same here since z-indices are always going to be application-dependent, and you could still see layering glitches if the dropdown is housed inside yet another stacking context.

I do see that replacing the `999` with a custom property helps, as applications could customize the z-index if they really need to, but I'd like to return to this solution only if portals don't appear to be an option.

Half-Shot:
Agreed with your comments, I'll have a go.

Half-Shot:
I don't think I'm likely to get to this one.

PR #280, opened 2024-11-19
Allow dropdown state to be controlled

robintown:
Controlling the dropdown value via its props is often more convenient and enables the value to be changed without user interaction.

### DIFF
--- a/src/components/Dropdown/Dropdown.test.tsx
+++ b/src/components/Dropdown/Dropdown.test.tsx
@@ -17,10 +17,10 @@ limitations under the License.
 import { describe, expect, it } from "vitest";
 import { composeStories } from "@storybook/react";
 import * as stories from "./Dropdown.stories";
-import { act, render, screen, waitFor } from "@testing-library/react";
+import { render, screen } from "@testing-library/react";
 import React, { FC, useMemo, useState } from "react";
-import { userEvent } from "@storybook/test";
 import { Dropdown } from "./Dropdown";
+import userEvent from "@testing-library/user-event";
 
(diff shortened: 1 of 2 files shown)


src/components/Dropdown/Dropdown.test.tsx
```diff
@@ -99,4 +120,20 @@ describe("Dropdown", () => {
...
   });
+  it("supports controlled operation", async () => {
+    const user = userEvent.setup();
+    render(<ControlledDropdown />);
+
+    expect(screen.getByRole("option", { name: "Option 1" })).toHaveAttribute(
+      "aria-selected",
+      "true",
+    );
+    await act(() =>
```
dbkr, line 131: My understanding of user-event is that it handles all the `act()` stuff internally so it doesn't need to be wrapped in act, instead you just await on user.click() itself, although I'm failing to find any specific docs other than that that's what the examples do: https://testing-library.com/docs/user-event/intro/

robintown: True :) now seems like a good time to figure out why this test suite spews act warnings without it

t3chguy: iirc in EW we had to add `globalThis.IS_REACT_ACT_ENVIRONMENT = true;` to the test setup file

robintown: I recall in the past this has been caused by the versions of testing-library dependencies being out of sync.

dbkr (approved):

PR #292, opened 2024-12-09
Add breadcrumb component

florianduros:
[Figma design](https://www.figma.com/design/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?node-id=3646-61594&node-type=frame&t=g1P0k26KwfXtgLLs-0)

To avoid the `/` separator to be read by a screen reader, I have multiple solutions:
- Add a `<span aria-hidden={true}>/</span>`. => I don't like to use an extra span for the separator, to keep the dom simple
- Use the alternative text `content: "/" / ""` instead of the border.  => In the accessiblity tree of Chrome, the text is still appearing as `static text: ""`
- Use a rotated border


florianduros:
@dbkr Not so bad but a little pain for screen readers :) It's easy to avoid

PR #307, opened 2025-01-17
Update License: AGPL + Commercial license

langleyd:
Update License: AGPL + Commercial license


src/components/Avatar/avatar-clip.mask.svg
dbkr, line 1: how did this one change?

src/components/Form/Controls/Action/Action.stories.tsx
```diff
@@ -1,18 +1,10 @@
 /*
+Copyright 2025 New Vector Ltd.
```
dbkr, line 2: This one seems to have gone weird, and the same thing seems to have happened to others.

langleyd: The tool we use to update it, sorts the holder by year descending. If the holder changes I believe we are meant update the year. As they are both equal I'm not sure it's really determinable which was the holder first in the existing format, although I think the most recent holders generally go to the top. In which case... I think this is correct? (albeit it might be more likely the change in holder actually went in the other direction)

I think it's probably good enough, vs me spending time special casing these edge cases across all the repos there are to update. WDYT?

dbkr: Oh right, that seems fair enough then.

src/components/Avatar/avatar-clip.mask.svg
langleyd: image

The header in the svg changed, press the source button in GH and you'll see it ^

dbkr: oh right, of course.

dbkr (approved):

PR #316, opened 2025-02-20
fix: tooltip and menu used together

florianduros:
Closes https://github.com/element-hq/compound/issues/333
Regression due to https://github.com/element-hq/compound-web/pull/290

We need to pass the children props otherwise tooltip and menu together don't work.

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -178,7 +178,8 @@ const TooltipAnchor: FC<TooltipAnchorProps> = ({
     if (!isValidElement(children)) return;
 
     if (isTriggerInteractive) {
-      const props = context.getReferenceProps({ ref });
+      // @ts-expect-error needed to fix https://github.com/element-hq/compound/issues/333
+      const props = context.getReferenceProps({ ref, ...children.props });
       return cloneElement(children, props);
     } else {
       // For a non-interactive trigger, we want most of the props to go on the


src/components/Tooltip/Tooltip.tsx
```diff
@@ -178,7 +178,8 @@ const TooltipAnchor: FC<TooltipAnchorProps> = ({
     if (!isValidElement(children)) return;
 
     if (isTriggerInteractive) {
-      const props = context.getReferenceProps({ ref });
+      // @ts-expect-error needed to fix https://github.com/element-hq/compound/issues/333
```
t3chguy, line 181: What is the error here?

florianduros: `TS2698: Spread types may only be created from object types.`

t3chguy: What is the type from `children.props`?
Is it specific to React <19?

florianduros: React 19: `children.props: unknown`
React 18: `children.props: any`

I can instead cast it to `children.props as object` to avoid the error

t3chguy: I think doing some actual type testing might be saner, `...(typeof children.props === "object" ? children.props : {})`

t3chguy: Props can be of any type, often you see a pattern of using function as children also.

florianduros: Oh, good idea

src/components/Tooltip/Tooltip.tsx
```diff
@@ -178,7 +178,11 @@ const TooltipAnchor: FC<TooltipAnchorProps> = ({
     if (!isValidElement(children)) return;
 
     if (isTriggerInteractive) {
-      const props = context.getReferenceProps({ ref });
+      const props = context.getReferenceProps({
+        ref,
+        // Needed to fix https://github.com/element-hq/compound/issues/333
+        ...(typeof children.props === "object" ? children.props : {}),
```
robintown, line 184: I'm unclear on why this works, for React 18. Why isn't it enough for `cloneElement` to merge these props? (Is the answer perhaps that `getReferenceProps` does some smarter merging of the event handlers it uses, internally?)

robintown (changes requested):
Though this may fix tooltips+menus on React 18, they remain broken on React 19 which includes `ref` in `children.props`. The menu functions but the tooltip does not appear. You can reproduce this with the story I posted here: https://github.com/element-hq/compound/issues/333#issuecomment-2618855511

robintown:
For React 19 support, I think it's enough to swap the order of the spread and `ref`:

```typescript
const props = context.getReferenceProps({
  ...(typeof children.props === "object" ? children.props : {}),
  ref,
});
```

florianduros:
Okay, your solution @robintown is working on react 18 too

robintown (commented):
I'd still appreciate a comment about why spreading props there works

robintown (approved):

PR #320, opened 2025-02-21
Add radio menu item

florianduros:
[Radio menu item in figma](https://www.figma.com/design/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?node-id=3733-18878&t=zubF29jviruPzhvT-4)

Implementation close to the `CheckboxMenuItem`


t3chguy:
@florianduros any idea why there is no cloudflare preview?

florianduros:
Hmm strange, I'll rebase on main to see if it's triggering a preview

src/components/Menu/RadioMenuItem.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+import React, { useState } from "react";
+import { RadioMenuItem as RadioMenuItemComponent } from "./RadioMenuItem.tsx";
+import { Meta, StoryObj } from "@storybook/react";
+
+type Props = Omit<
+  React.ComponentProps<typeof RadioMenuItemComponent>,
+  "Icon" | "label" | "checked" | "onSelect" | "name"
```
florianduros, line 14: Remove "Icon"

t3chguy:
This looks to not follow best wai-aria practices where arrow keys do not change the radio selection, only space/enter does https://ariakit.org/examples/menu-item-radio - this may just be a symptom of the Story. Could you update the story to match the best practice so that we can confirm the component can be used to achieve this

florianduros:
Catching up after a bit of time.

@t3chguy I'm looking at the wai aria pratices: https://www.w3.org/WAI/ARIA/apg/patterns/radio/
> Right Arrow and Down Arrow: move focus to the next radio button in the group, uncheck the previously focused button, and check the newly focused button. If focus is on the last button, focus moves to the first button.

Arrow keys should change the selection. (a radio group is missing in the story)

t3chguy:
> @t3chguy I'm looking at the wai aria pratices: [w3.org/WAI/ARIA/apg/patterns/radio](https://www.w3.org/WAI/ARIA/apg/patterns/radio/)

but that's wrong - `radio` != `menuitemradio` which is the role you're implementing.

dbkr:
ftr, https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-editor/ is an example of the menuitemradio best practice (although even that disagrees with https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menuitemradio_role on whether space should close the menu)

dbkr:
After much anguish, I think this is ready (we didn't need to manually control the keyboard focus at all, radix does it for us as long as we actually wrap the thing in a menu).

t3chguy (approved):

PR #322, opened 2025-03-04
Prevent PRs with the X-Blocked label from being merged

robintown:
Adapted from matrix-js-sdk (https://github.com/matrix-org/matrix-js-sdk/blob/develop/.github/workflows/pull_request.yaml)
For https://github.com/element-hq/voip-internal/issues/303

Needs the job to be configured as required for this to really *prevent* such PRs from being merged, but I'll take care of that shortly after this is merged.

### DIFF
--- a/.github/workflows/blocked.yml
+++ b/.github/workflows/blocked.yml
@@ -1,0 +1,17 @@
+name: Prevent blocked
+on:
+  pull_request:
+    types: [opened, labeled, unlabeled]
+jobs:
+  prevent-blocked:
+    name: Prevent blocked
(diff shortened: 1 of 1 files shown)


t3chguy:
It looks like no CI in this repo is actually marked as required, so this PR will not do what it says on the tin, it'll just show a red X but still allow merging. I lack sufficient permissions to change that.

robintown:
It turns out you need to list the specific status checks that should protect a branch, so I'm taking this out of draft; we should merge this *before* enabling branch protection.

dbkr (approved):

PR #324, opened 2025-03-12
Add onClick handler for `ToggleMenuItem`

florianduros:
`ToggleMenuItem` lacks the `onClick` handler. We can't for example prevent default this event.

### DIFF
--- a/src/components/Menu/MenuItem.tsx
+++ b/src/components/Menu/MenuItem.tsx
@@ -56,6 +56,11 @@ type Props<C extends MenuItemElement> = {
   // This prop is required because it's rare to not want a selection handler
   onSelect: ((e: Event) => void) | null;
   /**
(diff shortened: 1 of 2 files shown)


src/components/Menu/MenuItem.tsx
```diff
@@ -55,6 +55,11 @@ type Props<C extends MenuItemElement> = {
    */
   // This prop is required because it's rare to not want a selection handler
   onSelect: ((e: Event) => void) | null;
+  /**
+   * Event callback for when the item is clicked.
+   * @param e
```
dbkr, line 60: nit, but this line doesn't seem to be adding much - unless it's necessary for tsdoc to do the right thing, it seems a bit unnecessary.

dbkr (approved):

PR #326, opened 2025-03-13
fix(menu item): correct typing for onClick

florianduros:
This is embarrassing. 
I put `C` as generic of the `MouseEventHandler` which is obviously not an HTMLElement but a tag......

### DIFF
--- a/src/components/Menu/MenuItem.tsx
+++ b/src/components/Menu/MenuItem.tsx
@@ -60,7 +60,7 @@ type Props<C extends MenuItemElement> = {
    * Event callback for when the item is clicked.
    * @param e
    */
-  onClick?: MouseEventHandler<C>;
+  onClick?: MouseEventHandler;
   /**
    * The color variant of the menu item.
    * @default primary
--- a/src/components/Menu/MenuItem.tsx
+++ b/src/components/Menu/MenuItem.tsx
@@ -60,7 +60,7 @@ type Props<C extends MenuItemElement> = {
    * Event callback for when the item is clicked.
    * @param e
    */
-  onClick?: MouseEventHandler;
+  onClick?: MouseEventHandler<HTMLElementTagNameMap[C]>;
   /**
    * The color variant of the menu item.
    * @default primary


src/components/Menu/MenuItem.tsx
```diff
@@ -60,7 +60,7 @@ type Props<C extends MenuItemElement> = {
    * Event callback for when the item is clicked.
    * @param e
    */
-  onClick?: MouseEventHandler<C>;
+  onClick?: MouseEventHandler;
```
t3chguy, line 63:
```suggestion
  onClick?: MouseEventHandler<HTMLElementTagNameMap[C]>;
```

maybe? (untested)

t3chguy:
More importantly why didn't any type checks in this project catch it?

src/components/Menu/MenuItem.tsx
florianduros: It's working :)

TIL `HTMLElementTagNameMap`

t3chguy: We use it in AccessibleButton in element-web FTR

t3chguy:
I think its important to have checks to protect against such failures, what error did you see due to this before the fix? It'd be good to make sure type checks pick up on this.

florianduros:
In compound web, TS wasn't raising any error. `MouseEventHandler` accepts an 
```
Element extends Node, ARIAMixin, Animatable, ChildNode, NonDocumentTypeChildNode, ParentNode, Slottable
```
 where any string seems to satisfy *Element*.

But when using a `MenuItem` in EW, I had a mismatch type between: 
```
TS2345: Argument of type MouseEvent<"button", MouseEvent> is not assignable to parameter of type MouseEvent<Element, MouseEvent>
Type string is not assignable to type Element
```

t3chguy:
Right, that might be a React 18 vs 19 interplay. We might want CI in compound to test against both `@types/react` 18 & 19 due to it claiming to support both: https://github.com/element-hq/compound-web/blob/main/package.json#L121

Could you please file that as an issue? Once done I'm happy to approve this

florianduros:
Done in https://github.com/element-hq/compound-web/issues/327

t3chguy (approved):

PR #332, opened 2025-03-28
feat(menu item): make Icon optional

florianduros:


src/components/Menu/MenuItem.module.css
```diff
@@ -40,13 +56,6 @@ Please see LICENSE files in the repository root for full details.
   margin-inline-end: var(--cpd-space-4x);
 }
 
-.label {
```
florianduros, line 43: Moved because CSS linter want to be declared before:
```css
.item.no-icon {
    .label {}
} 
```

dbkr (approved):

PR #335, opened 2025-04-03
feat(avatar): add icon support

florianduros:
Add icon to avatar to support [this figma design](https://www.figma.com/design/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?node-id=4590-6096&t=LhqvkmJhdAIbrIs4-4)

The icon size can be changed independently of the avatar size with a default size of 16px (plus 2px of padding)

image


t3chguy:
This seems like it only works when rendered atop `canvas-bg` - anywhere else the outline around the icon will be the wrong colour to look like a gap.

t3chguy:
Also what about position? EW also uses similar cutouts on the top right of avatars, sometimes both at once. 

image

florianduros:
We can have a position parameter for the icon. 
Both position when?
(FYI, the collapsed room list is going away with the new room list)

t3chguy:
> Both position when?

Public with notification badge

> (FYI, the collapsed room list is going away with the new room list)

That sounds like it will cause a lot of backlash for vertical display users

florianduros:
> Public with notification badge

For public room when the room list is collapsed, right?

> That sounds like it will cause a lot of backlash for vertical display users

Maybe but this is a remark for the design team.

florianduros:
> This seems like it only works when rendered atop `canvas-bg` - anywhere else the outline around the icon will be the wrong colour to look like a gap.

This can be override by using the `iconClassName` props and specify the `background-color` in the class. This not the ideal but I wonder how I can easily have this empty gap between the avatar and icon.

t3chguy:
That won't work on gradients etc which need to match positionally. Like on the current room list and space panel for eg

florianduros:
sorry, I don't follow. You're talking about the indicators at the top right of the spaces?

t3chguy:
I mean in general mainly, these components should be highly reusable, not designed for one usage. Single use components should just be in the app layer.

florianduros:
The [room avatar](https://www.figma.com/design/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?node-id=4590-6096&t=WcE8w3c5384fP3SH-4) is in the compound figma. A good compromise is to create a RoomAvatar component in component built upon the Avatar component to implement strictly the RoomTypes.

florianduros:
Moving the implementation to EW.

PR #338, opened 2025-04-11
Remove non accessible link from the top of the readme

langleyd:
Remove non accessible link from the top of the readme


t3chguy (commented):
Feels weird to have a link which is inaccessible to the majority at the top of the README...

t3chguy (approved):

dbkr (approved):

PR #343, opened 2025-05-08
Add radio menu item (ARIA compliant version)

dbkr:
Based on https://github.com/element-hq/compound-web/pull/320 but implementing the ARIA menuitemradio behaviour described in https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menuitemradio_role

This is done by adding another container component which handles keyboard input manually to change focus which feels a bit ick but I couldn't see a better way.


dbkr:
Turns out this was not necessary: radix does all of this for us correctly, we just needed to wrap it in a Menu. Sigh.

PR #353, opened 2025-06-17
Add SettingsToggle component

Half-Shot:
This is a light implementation of the new settings toggle. It's mostly reusing existing components, but ensuring a consistent style across apps.

### DIFF
--- a/src/components/Form/Controls/SettingsToggle/SettingsToggle.tsx
+++ b/src/components/Form/Controls/SettingsToggle/SettingsToggle.tsx
@@ -1,0 +1,49 @@
+/*
+Copyright 2025 New Vector Ltd.
+Copyright 2023 The Matrix.org Foundation C.I.C.
+Copyright 2023 New Vector Ltd
+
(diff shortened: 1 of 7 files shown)


src/components/Form/Controls/SettingsToggle/SettingsToggle.tsx
```diff
@@ -0,0 +1,79 @@
...
+      control={<ToggleInput ref={ref} {...props} />}
+    >
+      <Label>{label}</Label>
+      {helpMessage && <HelpMessage>{helpMessage}</HelpMessage>}
+      {disabledMessage && props.disabled && (
+        <HelpMessage>{disabledMessage}</HelpMessage>
+      )}
+    </InlineField>
+  );
+  return content;
```
sandhose, line 64: Sorry to complain after it being merged, but I'm failing to see the value of that kind of component 😅 

I'm really in favour of using composition in React rather than defining everything as props. This is how you end up with components with 100s of props.

All the base components here already existed to support this exact use case, so why the need to define a *very specific component* here?

PR #377, opened 2025-08-14
Improve accessibility of toggle and checkbox menu items

robintown:
MenuItems can be used either inside a Menu component (which creates a roving tab index) or simply with an element of role `menu`. In the latter case, focus would previously land on the cosmetic `<input>` element, which is a bad idea as it's hidden from accessibility technologies. By turning the parent `<div role="menuitemcheckbox">` into a `<button>` instead, we can ensure that focus always lands on the parent element. This is possible because ToggleMenuItem and CheckboxMenuItem are controlled components these days.


t3chguy:
@robintown can you make CI happy?

t3chguy (approved):
LGTM, one thing I noticed (not affected by this PR) is disabled items are not reachable by keyboard so disabled items are almost invisible to non-sighted users, if you're still in the area of improving accessibility that'd be a good one to resolve

PR #420, opened 2026-01-02
Switch to NPM trusted publishing

Half-Shot:
For https://gh.io/all-npm-classic-tokens-revoked etc, we need to move to using trusted publishing. I've just copied the workflows we have for other repositories.

### DIFF
--- a/.github/workflows/npm_release.yml
+++ b/.github/workflows/npm_release.yml
@@ -52,6 +52,7 @@ jobs:
         id: npm-publish
         run: |
           npm publish --access public --provenance
+          echo "version=$(jq -r '.version' package.json)" >> $GITHUB_OUTPUT
 
       - name: 🧬 Create release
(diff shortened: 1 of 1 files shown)


.github/workflows/npm_release.yml
```diff
@@ -40,12 +47,11 @@ jobs:
...
           git push
+
       - name: 🚀 Publish to npm
         id: npm-publish
-        uses: JS-DevTools/npm-publish@7f8fe47b3bea1be0c3aec2b717c5ec1f3e03410b # v4
-        with:
-          token: ${{ secrets.ELEMENT_NPM_TOKEN }}
-          access: public
+        run: |
+          npm publish --access public --provenance
```
t3chguy: This won't work, L61-L63 refers to outputs from this which no longer exist

PR #482, opened 2026-04-16
Add Switch component

Half-Shot:
Pulled out of https://github.com/element-hq/element-call/blob/6d14f1d06ff007ce8a98f8e4e467fbb650da3796/src/room/LayoutToggle.tsx#L1

Provisional design from https://www.figma.com/design/NKjLzSlbD8JnldYQ5sFpce/Element-Call-Updates---Q1-2025?node-id=358-65788&t=sc0fU5HeJ2cnRSjL-4


src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+          type="radio"
+          aria-label={rightLabel}
+          name={id}
+          value={rightValue}
+          checked={value === rightValue}
+          onChange={onChange}
+        />
+      </Tooltip>
+      <RightIcon aria-hidden width={size} height={size} />
+    </div>
```
toger5: We recently changed this to be a from (instead of a div)
`<form className={classNames(styles.toggle, className)}>`
The reason is, that otherwise it will mess up the radio options input element if there are multiple toggle buttons on screen (like in the story for the bottom bar)
I think we should also do this in the shared component

src/components/ToggleButton/ToggleButton.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+  PlaySolidIcon,
+  StopSolidIcon,
+} from "@vector-im/compound-design-tokens/assets/web/icons";
+import { useState } from "storybook/internal/preview-api";
+
+const meta = {
+  title: "ToggleButton",
+  component: ToggleButton,
+  tags: ["autodocs"],
+  argTypes: {},
```
toger5: empty arg types? We can remove it all together?

toger5 (commented):
Sorry for reviewing while it not yet ready to review.

robintown:
@Half-Shot Could you help me understand the use case for this outside of Element Call? This repository should only house components that belong to the Compound Web Components library in Figma, so I wanna make sure that the design team is on board with upstreaming it.

src/components/ToggleButton/ToggleButton.tsx
Half-Shot: I had originally copied it verbatim, but then had issues getting this working at different sizes. I will go back and try again.

playwright/visual.test.ts-snapshots/Switch-Default-1-contrast-linux.png
robintown, line 1: Looks like there's nothing visually distinguishing the selected icon in high contrast mode

src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,76 @@
...
+  label: string;
+};
+
+interface Props<LeftValue extends string, RightValue extends string> {
+  left: Button<LeftValue>;
+  right: Button<RightValue>;
+  value: LeftValue | RightValue;
+  className?: string;
+  onChange: (value: LeftValue | RightValue) => void;
+  size?: "lg" | "md";
```
robintown:
```suggestion
  size?: Size & ("md" | "lg");
```

src/components/Switch/Switch.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { render } from "@testing-library/react";
+import React from "react";
+import { composeStories } from "@storybook/react";
+
+import * as stories from "./Switch.stories";
+import userEvent from "@testing-library/user-event";
+
+const { Default } = composeStories(stories);
+
+describe("Switch", () => {
```
robintown, line 18: For accessibility it would be reassuring to have tests that:

- The correct radio control is `checked`
- Keyboard interaction works

src/components/Switch/Switch.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+  it("renders", () => {
+    const { asFragment } = render(<Default />);
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it("can be toggled", async () => {
+    const user = userEvent.setup();
+    const onChange = vi.fn();
+    const { getByLabelText } = render(<Default onChange={onChange} />);
+    await user.click(getByLabelText("Stop"));
```
robintown:
```suggestion
    await user.click(getByRole("radio", { name: "Stop" }));
```

src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,76 @@
...
+  size = "lg",
+}: Props<LeftValue, RightValue>): React.ReactElement => {
+  const onInputChange = useCallback(
+    (e: ChangeEvent<HTMLInputElement>) =>
+      onChange(e.target.value as LeftValue | RightValue),
+    [onChange],
+  );
+
+  return (
+    <form data-size={size} className={classNames(styles.toggle, className)}>
```
robintown: The preferred pattern for \<input\> components is to require the consumer to manually wrap them in a `<Form>`, that way they can be composed into a larger \<form\> if necessary.

src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,76 @@
...
+          value={left.value}
+          checked={value === left.value}
+          onChange={onInputChange}
+        />
+      </Tooltip>
+      <left.Icon aria-hidden />
+      <Tooltip label={right.label}>
+        <input
+          type="radio"
+          name={right.value}
```
robintown: The `name` props of these two inputs should match so that they form a proper radio group. That can come from a required `name` prop on the Switch component, like it does in the case of our other \<input\> components

src/components/Switch/Switch.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+    left: {
+      Icon: PlaySolidIcon,
+      label: "Play",
+      value: "play",
+    },
+    right: {
+      Icon: StopSolidIcon,
+      label: "Stop",
+      value: "stop",
+    },
```
robintown: I think it would make sense to use the list & grid icons from Figma here to represent the most expected use case of the component. Bonus points if you could make the icons pickable from a dropdown like they are here: https://github.com/element-hq/compound-web/blob/3037f4d7c1d501c895aa8aa66f38effa7762b532/src/components/Button/Button.stories.tsx#L44

src/components/Switch/Switch.module.css
```diff
@@ -0,0 +1,102 @@
...
+    --cpd-switch-icon-padding: 10px;
+  }
+
+  &[data-size="md"] {
+    --cpd-switch-button-size: 32px;
+    --cpd-switch-icon-size: 20px;
+    --cpd-icon-button-margin: 0px;
+  }
+
+  padding: var(--cpd-space-1x);
```
robintown: This is more padding than the designs show (2px)

robintown (changes requested):

playwright/visual.test.ts-snapshots/Switch-Default-1-contrast-linux.png
Half-Shot: Odd, I've not run into this before. On storybook itself it seems to be working as intended?

image

robintown: Ah I could've been more specific, the "light-high-contrast" option you see in Storybook refers to the high contrast Compound theme, whereas the `contrast` in these file names actually refers to [forced colors mode](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/forced-colors)

robintown: I think you'll have to turn that on in your browser or system settings if you want to test it.

src/components/Switch/Switch.tsx
Half-Shot: Agreed. Wrapped the buttons in a `<fieldset>` and will require consumers to use a `<form>`.

src/components/Switch/Switch.tsx
Half-Shot: Yeah, that was dumb :laughing:

playwright/visual.test.ts-snapshots/Switch-Default-1-contrast-linux.png
Half-Shot: And test we did! Looks like it needs more borders.

src/components/Switch/Switch.module.css
```diff
@@ -0,0 +1,113 @@
...
+  border-radius: var(--cpd-radius-pill-effect);
+  background: var(--cpd-color-bg-canvas-default);
+  display: flex;
+  position: relative;
+  inline-size: fit-content;
+  border: 1px solid var(--cpd-color-border-interactive-secondary);
+
+  @media (forced-colors) {
+    border: 1px solid;
+  }
```
t3chguy, line 31: Confused as to how this adds anything given there is already a 1px border?

src/components/Switch/Switch.module.css
```diff
@@ -0,0 +1,113 @@
...
+    cursor: pointer;
+    border-radius: var(--cpd-radius-pill-effect);
+    background: var(--cpd-color-bg-action-secondary-rest);
+    box-shadow: 0 1.2px 2.4px 0 rgb(0 0 0 / 15%);
+    transition: background-color 0.1s;
+    border: 1px solid var(--cpd-color-bg-action-secondary-pressed);
+
+    @media (forced-colors) {
+      border: 1px solid;
+    }
```
t3chguy: Ditto

src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,81 @@
+/*
+Copyright (C) 2026 Element Creations Ltd
```
t3chguy: This looks different to the normal copyright in this repo `(C)`

t3chguy (approved):
Looks sane otherwise

src/components/Switch/Switch.module.css
```diff
@@ -66,6 +68,11 @@ Please see LICENSE in the repository root for full details.
     background: var(--cpd-color-bg-action-secondary-rest);
     box-shadow: 0 1.2px 2.4px 0 rgb(0 0 0 / 15%);
     transition: background-color 0.1s;
+    border: 1px solid var(--cpd-color-bg-action-secondary-pressed);
```
robintown: This should use a border token - looks like the designs have `--cpd-color-border-interactive-secondary` becoming `--cpd-color-border-interactive-primary` when hovered/pressed.

src/components/Switch/Switch.tsx
```diff
@@ -50,27 +55,27 @@ export const Switch = <
   );
 
   return (
-    <form data-size={size} className={classNames(styles.toggle, className)}>
-      <Tooltip label={left.label}>
+    <fieldset data-size={size} className={classNames(styles.toggle, className)}>
```
robintown: Firefox accessibility tools tell me:

Screenshot From 2026-04-24 10-29-32

I would be fine with either this component accepting a mandatory `aria-label` value or simply becoming a div.

src/components/Switch/Switch.module.css
Half-Shot: Without:

image

With:

image

I don't know why, it seems to be a Firefox-ism. I will add a comment.

src/components/Switch/Switch.module.css
Half-Shot: This one *is* unnecessary though

src/components/Switch/Switch.tsx
Half-Shot: I've opted to just expose the fieldset attributes wholesale so the user can make that choice.

robintown (approved):
Looks great, thanks!

PR #483, opened 2026-04-16
Use stable identifiers for generated CSS class names

richvdh:
Currently, there is typically a lot of work to do after upgrading between compound-web releases, in updating test snapshots. This is because CSS classes are named based on a hash of the class module's *contents*.

Currently, classes have names like `._text-content_zx76t_44`, where `text-button` is the original css name, `zx76t` is a hash of the css module file, and `44` is the line number within the file.

We switch to names like `_cpd-src-components-Alert-_text-content`, where `src-components-Alert-` is the path to the CSS module (with `/` replaced by `-`).

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -78,6 +78,15 @@ export default defineConfig({
     }),
   ],
 
+  css: {
+    modules: {
+      // Attempt to give generated CSS class names relatively stable identifiers.
+      // By default they include a hash of the css module file, and the line number within the file.
+      // Instead, let's just use the filename of the component for disambiguation.
+      generateScopedName: "_cpd-[path]_[local]",
+    },
+  },
+
   test: {
     environment: "jsdom",
     globals: true,


t3chguy:
We support loading multiple Compound versions at once, e.g. in Element Modules, and the styles should not conflict between them, this would cause that as you lose any versioning (provided by the hash).

This change would cause components to render incorrectly as the React component would match the version loaded and expected by the module, but the styles would conflict with that of Element Web's version of Compound Web, or worse yet depending on load order, the module could break Element Web's rendering of Compound components.

Currently this would impact the following modules which we maintain:
+ content-scanner
+ multiroom
+ banner
+ restricted-guests
+ widget-toggles

Might be possible work around by through intensive use of the shadow DOM but would be vastly non-trivial and add a lot more complexity to the implementation of the Module API

richvdh:
> We support loading multiple Compound versions at once, 

Ugh. That's annoying. Ok, thanks for the feedback.

PR #486, opened 2026-04-21
Import useId directly from React

dbkr:
As per the eslint message, floating-ui's useId wrapper messes up mocking of useId that's useful for making it return consistent IDs in tests.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -19,13 +19,19 @@ module.exports = {
   },
   rules: {
     "prettier/prettier": "error",
-    "no-restricted-imports": ["error", {
-      "paths": [{
-        "name": "@floating-ui/react",
-        "importNames": ["useId"],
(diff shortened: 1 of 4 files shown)


.eslintrc.cjs
```diff
@@ -19,6 +19,13 @@ module.exports = {
   },
   rules: {
     "prettier/prettier": "error",
+    "no-restricted-imports": ["error", {
+      "paths": [{
+        "name": "@floating-ui/react",
+        "importNames": ["useId"],
+        "message": "Please use useId from React directly: compound only support React 18+ and floating-ui's wrapper breaks mocks that make the IDs consistent in tests.",
```
t3chguy:
```suggestion
        "message": "Please use useId from React directly: Compound only supports React 18+ and floating-ui's wrapper breaks mocks that make the IDs consistent in tests.",
```

t3chguy (approved):

PR #506, opened 2026-06-06
feat(Tooltip): add boundary prop to hide tooltip when it escapes an element

Tamajit-005:
## Summary

Adds an optional `boundary` prop to `useTooltip` (and therefore `Tooltip`) that activates FloatingUI's `hide` middleware with `strategy: "escaped"`. When the tooltip would overflow the provided boundary, the floating element receives the existing `invisible` class (`clip-path: inset(50%)` + `pointer-events: none`), allowing it to remain in the accessibility tree while being visually hidden.

The boundary is also forwarded to the `flip` middleware so the initial placement is more likely to remain within the same boundary.

The prop is fully backward compatible. When `boundary` is omitted, tooltip behavior remains unchanged.

## Motivation

Element Web's "Expand map" tooltip on location messages can overlap the message composer at the bottom of the timeline. Constraining the tooltip to the message panel boundary allows it to hide cleanly when it would escape that area, without requiring consumers to implement their own FloatingUI middleware configuration.

## Implementation

`useTooltip.ts` adds `boundary?: Element | null | Element[]` to `CommonUseTooltipProps`, forwards it to the existing `flip` middleware, and conditionally appends `hide({ strategy: "escaped", boundary, padding: 6 })` to the middleware list. `Tooltip.tsx` reads `middlewareData?.hide?.escaped` and applies the existing `invisible` class when true. No CSS change is needed.

## Follow-up

A corresponding Element Web PR will consume this prop on `MLocationBody` to keep the "Expand map" tooltip from overlapping the message composer. That PR will land in element-hq/element-web once this change is merged and a new `@vector-im/compound-web` version is published.

### DIFF
--- a/src/components/Tooltip/useTooltip.ts
+++ b/src/components/Tooltip/useTooltip.ts
@@ -87,7 +87,7 @@ export interface CommonUseTooltipProps {
    * near containers they should not overlap (e.g. the message composer).
    * @default undefined
    */
-  boundary?: Element | null | Element[];
+  boundary?: Element | Element[];
 
   /**
    * Additional aria-* attributes to pass through to the floating tooltip for
@@ -156,19 +156,14 @@ export function useTooltip({
         crossAxis: placement.includes("-"),
         fallbackAxisSideDirection: "start",
         padding: 5,
-        // Only pass boundary if set — FloatingUI's Boundary type excludes null
-        ...(boundary ? { boundary } : {}),
+        boundary,
       }),
       shift({ padding: 5 }),
-      ...(boundary
-        ? [
-            hide({
-              strategy: "escaped",
-              boundary,
-              padding: 6,
-            }),
-          ]
-        : []),
+      hide({
+        strategy: "escaped",
+        boundary,
+        padding: 6,
+      }),
       // add the little arrow along with the floating content
       arrow({
         element: arrowRef,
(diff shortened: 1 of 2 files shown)


src/components/Tooltip/useTooltip.ts
```diff
@@ -77,6 +78,17 @@ export interface CommonUseTooltipProps {
...
   isTriggerInteractive: boolean;
 
+  /**
+   * An optional boundary element to constrain the tooltip within.
+   * When provided, the tooltip will be hidden if it escapes this boundary.
+   * Also passed to the `flip` middleware so it can choose a placement that
+   * keeps the tooltip inside the boundary.
+   * Accepts a single Element or an array of Elements. Useful for tooltips
+   * near containers they should not overlap (e.g. the message composer).
+   * @default undefined
```
t3chguy, line 88: what is the difference between `null` and `undefined` values being provided?

src/components/Tooltip/useTooltip.ts
```diff
@@ -143,8 +156,19 @@ export function useTooltip({
         crossAxis: placement.includes("-"),
         fallbackAxisSideDirection: "start",
         padding: 5,
+        // Only pass boundary if set — FloatingUI's Boundary type excludes null
+        ...(boundary ? { boundary } : {}),
```
t3chguy: why not use `boundary: boundary ?? undefined` then?

Tamajit-005: Fixed both. 
Dropped null from the type and simplified the middleware.

t3chguy (approved):
Looks sane to me, thanks

PR #510, opened 2026-06-16
Split out the menu part of dropdown

dbkr:
To allow it to be used separately from Dropdown itself, as we now have designs that require this.

It could also probably be unified a bit with other menu components, this splits it out to make this possible.


dbkr:
Going to achieve what I need (rendering the set status view in element web / shared components) with https://github.com/element-hq/compound-web/pull/511 instead so I don't need this.

PR #511, opened 2026-06-17
Allow dropdown to have a customisable trigger

dbkr:
Adds a customTrigger prop to dropdown that is called to render a trigger element instead of the standard box with selected option & chevron.

If a custom trigger is supplied, the menu floats just below it with all-rounded borders. With the standard trigger, it appears to extend up behind the trigger as it did before, although this PR removes the 'border' element that was used to render the small continuation of the border between the menu and the trigger. It replaces it with some CSS to extend the top element up slightly with padding, offset the menu upwards by the same amount and z-index the trigger so the menu hides underneath.

Removing the border element could be done in a separate PR but in practice I've worked on the two together, but I can split out if desired.

Screenshot 2026-06-18 at 13 56 20


playwright/components/Dropdown/Dropdown.test.ts-snapshots/arrow-down-contrast-linux.png
dbkr, line 1: Updated because the bottom of the blue a11y border was previously hidden under the menu, which is now fixed.

playwright/components/Dropdown/Dropdown.test.ts-snapshots/open-default-chromium-linux.png
dbkr, line 1: Updated because the menu's drop shadow was on top of the trigger, now fixed.

playwright/components/Dropdown/Dropdown.test.ts-snapshots/open-option-2-chromium-linux.png
dbkr, line 1: As above

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -91,3 +91,20 @@ export const WithDefaultValue: Story = {
...
+      <button aria-label="Custom trigger" {...props}>
+        🐴
+      </button>
+    ),
+    label: null,
+  },
+  parameters: {
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/file/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?node-id=799-5732&t=g2Ex9sbzgku1nTIN-4",
```
t3chguy: I think I'm being blind, where in this link does it show a custom trigger? I see only one type of trigger:

`[optional icon] [text] \/`

dbkr: Ah true, I just copied from the default but the custom trigger is entirely for the control to set user status - maybe it's clearer not to link to the figma here.

t3chguy: I'm of the opinion compound should only support things in the compound figma library, anything else is scope creep.

dbkr: As in you think it should be added to the figma or that we should solve the problem a different way?

t3chguy: Yes, either or

playwright/visual.test.ts-snapshots/Dropdown-With-Custom-Trigger-1-chromium-linux.png
robintown, line 1: May we have a screenshot where the dropdown is open on a custom trigger?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -68,6 +85,12 @@ type DropdownProps = {
    * The error message to display.
    */
   error?: string;
+  /**
+   * Render a custom trigger element instead of the default button.
+   * Receives the accessibility and event handler props to apply to the trigger,
+   * and the current display text (selected value label or placeholder).
+   */
+  customTrigger?: (props: DropdownTriggerProps) => React.ReactNode;
```
robintown: Perhaps we could just call this thing `trigger` and tag it with `@default A standard combobox with a chevron showing the selected option.`

src/components/Dropdown/Dropdown.module.css
```diff
@@ -30,18 +30,26 @@ Please see LICENSE files in the repository root for full details.
...
     }
   }
 
+  .open-trigger {
+    /*
+     * When the dropdown is open, it needs to sit above the menu so
+     * the menu appears to extend out from under the trigger.
+     */
+    z-index: 1;
+  }
```
robintown: We've no usages of `z-index` in compound-web currently since the values are often arbitrary and application-specific. Is there another way to make this work with `z-index: auto`? (For instance reversing the order of the elements in the DOM? :grimacing:)

src/components/Dropdown/Dropdown.stories.tsx
dbkr: Will chat with Aaron when he's back on Monday.

playwright/visual.test.ts-snapshots/Dropdown-With-Custom-Trigger-1-chromium-linux.png
dbkr: I can't add a story since the open-ness isn't controlled by a prop. I could add a playwright test that takes a screenshot (I've added one to the PR description too).

src/components/Dropdown/Dropdown.tsx
dbkr: Done

src/components/Dropdown/Dropdown.module.css
dbkr: I've managed to do it by re-ordering them in the DOM and shunting the menu downwards to the bottom of the container to compensate, not 100% sure whether it's cleaner but it does avoid a z-index. wdyt?

robintown: It looks fine :) …but oh, of course now the issue is that the elements are in the opposite tab order from what you'd expect.

playwright/visual.test.ts-snapshots/Dropdown-With-Custom-Trigger-1-chromium-linux.png
robintown: OK, a Playwright test would be nice then

robintown (commented):
For me, the corners of the first element disappear when I hover it:

Screenshot From 2026-06-18 16-03-25

src/components/Dropdown/Dropdown.module.css
robintown: This video shows me opening the dropdown with Space, pressing Shift+Tab, and focus landing on the last option in the dropdown:

[Screencast From 2026-06-18 16-05-43.webm](https://github.com/user-attachments/assets/7d4646eb-46f4-4518-b6c0-019c6b86432f)

playwright/visual.test.ts-snapshots/Dropdown-With-Custom-Trigger-1-chromium-linux.png
dbkr: Done!

src/components/Dropdown/Dropdown.module.css
dbkr: Hmm, yes. We could fix this with `tabIndex` but this will interact with other `tabIndex`ed things in the DOM in the same way as `z-index` so it feels like it defeats the whole point. At least with a z-index we can `position: relative` the container to establish a completely separate stacking context. Only other option I can think of is to disable tab entirely: after all, a native html select would just use up/down to navigate and tab would do nothing, I think?

robintown: > At least with a z-index we can `position: relative` the container to establish a completely separate stacking context.

Ah, are you saying the z-index could be application agnostic that way? I admit I don't fully grok stacking contexts…

> Only other option I can think of is to disable tab entirely: after all, a native html select would just use up/down to navigate and tab would do nothing, I think?

This sounds reasonable, though @t3chguy maybe you could offer your opinion on this as well? For me a native select element moves focus into the selection list automatically and also closes the moment I tab to something else, so there's probably quite a lot that's eligible for change in our Compound implementation about the tab & focus behavior.

Ultimately a non-`auto` z-index is okay as long as it can be customized, just wanted to explore the other options first.

dbkr: > > At least with a z-index we can `position: relative` the container to establish a completely separate stacking context.
> 
> Ah, are you saying the z-index could be application agnostic that way? I admit I don't fully grok stacking contexts…

Yep, in fact I just did some further research and it turns out [`isolation: isolate`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/isolation) was created for this purpose (ie. just create a new stacking context and nothing else) so this is feeling like the best option to me.

dbkr:
I've forked off another branch at the point before I started changing to the dom-order method and put that in https://github.com/element-hq/compound-web/pull/512 adding isolation:isolate (and porting the extra tests/fixes in) as reverting the commits seems awful and force pushing will lose everything. I've assigned you two manually as reviewers.

dbkr:
Replaced by https://github.com/element-hq/compound-web/pull/512

PR #512, opened 2026-06-22
Allow dropdown to have a customisable trigger: mk II

dbkr:
Adds a customTrigger prop to dropdown that is called to render a trigger element instead of the standard box with selected option & chevron.

If a custom trigger is supplied, the menu floats just below it with all-rounded borders. With the standard trigger, it appears to extend up behind the trigger as it did before, although this PR removes the 'border' element that was used to render the small continuation of the border between the menu and the trigger. It replaces it with some CSS to extend the top element up slightly with padding, offset the menu upwards by the same amount and z-index the trigger so the menu hides underneath.

Removing the border element could be done in a separate PR but in practice I've worked on the two together, but I can split out if desired.

Alternative impl to https://github.com/element-hq/compound-web/pull/511 (using z-index and isolation: isolate)


americanrefugee (approved):
LGTM

dbkr:
@t3chguy @robintown does this seem good to you?

t3chguy (approved):

PR #519, opened 2026-07-01
Migrate to pnpm

florianduros:
Replaces Yarn with pnpm as the package manager.

- **Package manager**: Added `"packageManager": "pnpm@11.9.0"` to `package.json` and replaced `yarn.lock` with `pnpm-lock.yaml`
- **CI workflows**: Updated all GitHub Actions workflows (`analysis.yml`, `npm_release.yml`, `storybook.yml`) to use `pnpm/action-setup` and pnpm commands
- **Scripts**: Updated all `yarn …` references in `package.json` scripts to `pnpm …`
- **Workspace config**: Added `pnpm-workspace.yaml` with an `allowBuilds` allowlist to control native addon compilation
- **Missing deps**: Added `@storybook/react` and `@floating-ui/react-dom` which were implicit transitive dependencies not declared directly
- **README**: Updated command examples to use pnpm; removed the `yarn link` workaround (no longer applicable with pnpm's symlink-based linking)
- **Prettier**: Added `pnpm-lock.yaml` to `.prettierignore`

> [!CAUTION]
> PR should be merged at the same time than the cloudflare settings update.
> [Cloudflare settings](https://gitlab.matrix.org/infrastructure/cloudflare/-/blob/main/pages/compound-web.tf?ref_type=heads#L7) should be updated to use `pnpm install --frozen-lockfile --ignore-scripts && pnpm build-storybook` to deploy storybook.


.github/workflows/npm_release.yml
```diff
@@ -25,20 +25,23 @@ jobs:
...
           node-version-file: ".node-version"
 
       - name: 🛠️ Setup
-        # When running `install` it also calls the `prepare` step which generates
-        # a build
-        run: yarn install --pure-lockfile
+        run: pnpm install --frozen-lockfile --ignore-scripts
+
+      - name: 🚧 Build
+        run: pnpm build
```
dbkr, line 40: sonar is probably right to want ignore scripts here (and elsewhere)

package.json
```diff
@@ -60,6 +61,7 @@
     "@storybook/addon-docs": "^10.3.1",
     "@storybook/addon-links": "^10.3.1",
     "@storybook/addon-themes": "^10.3.1",
+    "@storybook/react": "^10.3.1",
```
dbkr, line 64: why did we gain dependencies?

dbkr (commented):
Thanks for doing this!

package.json
florianduros: pnpm and yarn don't organize the deps at the same level. We are using these deps without declaring them (they are deps of other deps) and moving to pnpm flaged it.

.github/workflows/npm_release.yml
florianduros: Yes, a sane approach

package.json
dbkr: ah okay, cool

.github/workflows/npm_release.yml
t3chguy: pnpm has a whole `allowBuilds` thing where you allowlist scripts on a per-dep basis and Sonar doesn't consider it

dbkr (approved):

PR #523, opened 2026-07-07
Make Link component feel more like a link

robintown:
Aaron and I decided to tweak the styles of our Link component a little further to help it actually "feel" like a link during interactions.

[Screencast From 2026-07-07 17-44-25.webm](https://github.com/user-attachments/assets/fc57a860-c621-4667-b497-c8d77a023cc2)


t3chguy (changes requested):
Should probably have a styling for prefers-reduced-motion

robintown:
I don't normally add prefers-reduced-motion rules for animations that are purely color/opacity based, since there's presumably no real motion there that could be a vestibular trigger. (I wrote as much in the [guidelines](https://compound.element.io/?path=/docs/develop-contributing--docs)) Button and Switch are the same way with their transitions.

t3chguy (approved):

PR #524, opened 2026-07-09
Refine the design of Switch

robintown:
The [designs](https://www.figma.com/design/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?node-id=12554-2698&t=Zx6desTK2XigACOj-4) have changed to use a subtle background instead of an outline, which hopefully looks a bit cleaner. I also took the opportunity to get the size of the component pixel-perfect.


playwright/Dockerfile
t3chguy, line 1: This change seems unrelated?

robintown: Oops, yes. The dockerfile has been broken since the migration to pnpm. I meant to put this in a separate PR.

robintown: Force pushed to remove it

t3chguy (approved):

PR #526, opened 2026-07-09
Fix active Switch icon being invisible in forced colors mode

robintown:
Sorry, hit one more issue with the Switch component when I finally managed to activate forced colors on Chromium…

|Before|After|
|-|-|
|Screenshot From 2026-07-09 18-30-28|Screenshot From 2026-07-09 18-30-18|

### DIFF
--- a/src/components/Switch/Switch.module.css
+++ b/src/components/Switch/Switch.module.css
@@ -108,7 +108,9 @@ Please see LICENSE in the repository root for full details.
   }
 
   input:checked + svg {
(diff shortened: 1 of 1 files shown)


t3chguy:
Any clue why this wasn't caught in CI?

robintown:
Not really… it is of course Chromium in both cases 😕

robintown:
Oh, I'm wrong. The forced colors screenshots are taken in Firefox.

t3chguy (approved):

PR #531, opened 2026-07-22
Change menu max width to 340px

dbkr:
Matches the menu width in compound and needed because the emoji picker is 340px


florianduros (approved):

americanrefugee (approved):
LGTM

t3chguy (approved):

PR #538, opened 2026-08-06
Change Dropdown to have a renderItem callback

dbkr:
Allows more customisation of what gets rendered. Also make it type parameterized over the type of the values it can take so we can make stronger type assertions.

This is a breaking change although the only known usage of this component is the user status control I'm making the change for.

We could add a wrapper on top to take a simple array for the values as before, but the callback is so trivial to provide it hardly seems worth it.

Also update to use react 19 ref prop rather than forwardRef. Unfortunately this does change the indenting and therefore make the diff horrendous.

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -198,7 +198,7 @@ export function Dropdown<K extends string | number = string>({
           ref={combinedRef}
           {...props}
         >
-          {currentContent}
+          <div>{currentContent}</div>
           <ChevronDown className={styles.chevron} width="24" height="24" />
         </button>
       )}
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -44,7 +44,7 @@ export type DropdownTriggerProps = {
   onKeyDown?: (e: KeyboardEvent<Element>) => void;
 };
 
-export type DropdownProps<K extends string | number = string> = {
+type DropdownProps<K extends string | number = string> = {
   /**
    * The CSS class name.
    */
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -44,7 +44,7 @@ export type DropdownTriggerProps = {
   onKeyDown?: (e: KeyboardEvent<Element>) => void;
 };
 
-type DropdownProps<K extends string | number = string> = {
+type DropdownProps<K = string> = {
   /**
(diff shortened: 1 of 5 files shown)


src/components/Dropdown/__snapshots__/Dropdown.test.tsx.snap
t3chguy, line 1: Does this mean it isn't possible to render without the spurious div like in the old behaviour?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -42,30 +44,25 @@ export type DropdownTriggerProps = {
   onKeyDown?: (e: KeyboardEvent<Element>) => void;
 };
 
-type DropdownProps = {
+type DropdownProps<K extends string | number = string> = {
```
t3chguy: Why so limiting? What if I want to pass an array of objects without needing to pre-format them so `renderItem` can access the fields

src/components/Dropdown/Dropdown.tsx
```diff
@@ -91,153 +88,153 @@ type DropdownProps = {
...
+      {trigger ? (
+        trigger(triggerProps)
+      ) : (
+        <button
+          className={buttonClasses}
+          {...triggerProps}
+          ref={combinedRef}
+          {...props}
+        >
+          <div>{currentContent}</div>
```
t3chguy, line 201: Why the extra wrapping div?

dbkr: The parent is a flex container that puts the arrow on the right hand side so this keeps the actual content as one item in the flex container. We could leave it out and leave the renderItem callback to add a wrapper if it renders more than one element (ie. not use a fragment), it would just be a bit of a footgun in the API.

src/components/Dropdown/__snapshots__/Dropdown.test.tsx.snap
dbkr: It does, see other comment.

src/components/Dropdown/Dropdown.tsx
dbkr: Yeah, would be a little weird to receive it back in the renderItem callback, but on the other hand I guess you could just define it as a function of the array! Removed in 40b9cda0.

src/components/Dropdown/Dropdown.tsx
t3chguy: Semantically maybe should it be a span given it is inline (or inline-block at best)?

t3chguy (approved):

PR #544, opened 2026-08-20
Always spin the InlineSpinner

Half-Shot:
Stopping the spinner when reduced motion is turned on had the unfortunate side effect of making the application look broken when not animating.

Since this animation *is* semantic, I would vote we keep it given the actual animation is quite small / unintrusive.


robintown (commented):
Came across a possible nice compromise in the wild just now: What about slowing down the animation, [like Bootstrap does](https://getbootstrap.com/docs/5.1/getting-started/accessibility/#reduced-motion)?

Half-Shot:
Genius! Yeah let's do that

florianduros:
Should it back to draft?

PR #550, opened 2026-09-16
Add maxWidth and maxLines property to Tooltip

dbkr:
So consumers can limit the width of the tooltip and the number of lines it can be.

It might seem a bit silly to give Tooltip a bunch of text and then tell it to only display some of it rather than just give it less text, but this lets the width be limited by CSS allowing for limiting by pixels, percent, ch units etc rather than just bluntly chopping to a fixed number of characters (and having to worry about not chopping mid-grapheme cluster etc).

This uses the prefixed line-clamp which automatically adds an ellipsis. Tested in Chrome, Firefox and Safari.


florianduros (approved):

gaelledel (commented):
What size limit is this set to? Did you say the limit is configurable ?

dbkr:
I've not set a default so it won't affect anything until we enable it for each tooltip. I can do though if that would be useful.

gaelledel:
yes I wonder if we could set a max default? i.e: What I set up in this example https://www.figma.com/design/MAUsalKv7bRRNKlAAigtNd/Community-contributions---Product-Polish?node-id=188-21304&t=G33o8maP9gN3eipo-4

dbkr:
Added a default width of 580px (the figma says 579 but I figure it should at least be an even number, if not a multiple of 8) and 2 lines.

gaelledel (approved):

PR #551, opened 2026-09-17
Add disabled state to Dropdown

dbkr:
Screenshot 2026-09-17 at 18 15 36


t3chguy (commented):
The cursor when disabled remains pointer which seems misleading

dbkr:
True, I've swapped it for not-allowed, not sure whether we want that or default.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -188,7 +201,11 @@ export function Dropdown<K extends string | number = string>({
       className={classNames(className, styles.container)}
       aria-invalid={Boolean(error)}
     >
-      {label && <label id={labelId}>{label}</label>}
+      {label && (
+        <label id={labelId} className={classNames(disabled && styles.disabled)}>
```
t3chguy:
```suggestion
        <label id={labelId} className={classNames({ [styles.disabled]: disabled })}>
```

t3chguy (approved):

